PR #62, opened 2014-10-30
Suggested Korean Translation

JoshGraham:
I've ignored
<string tools:ignore="MissingTranslation" name=
because I didn't know whether to translate or not.
Will translate if needed.

Thank you.


Ph1b:
Dear Josh, thanks again for your translation!

I just created a [project page](https://www.transifex.com/projects/p/material-audiobook-player/) for translations. Do you want to be the maintainer of the korean language?
There are a few new strings that are still untranslated.

PR #123, opened 2015-01-23
fixes crash when book completes, closes #122

### LINKED ISSUE #122: MAP crashes after a book completes

dark-seizo:


Ph1b:
Sorry, but that did just fixed it partially, because the default media player (used in api < 16) cant be overwritten. So I tried a more general approach.
Any way, thanks for the contribution, keep the work up :) Ill merge the autoskip in the next release.

PR #127, opened 2015-01-27
Auto rewind after a pause

dark-seizo:
This is a nice feature I enjoyed having in my previous audiobook app.  It rewinds a small amount (2s + the log of the time paused in seconds) whenever the app is paused.  I found it useful when you come back to a book after a few hours, it would rewind ~10 seconds so to give you time to remember where you are, and what's going on.  It's also nice when your using GPS navigation, after each spoken direction it will back up ~2 seconds which minimizes the impact of the interruption.


Ph1b:
Hey dark-seizo, thanks for the pull request. 

The way you implemented the autorewind, the rewinding would just work on api > 16 because below, the default android MediaPlayer class is being used. So a more appropriate way would be to use the MediaPlayerCompat class.

PR #167, opened 2015-04-23
Fix incorrect playback speed (mono files) on some android devices

dreamdt:
Hello, Paul. 
Today i can't use player at all, most of my books are in mono.
I decided to send patch that fixes incorrect speed playback. You can rewrite it as you wish.
It's a safe way. I dont think audio file can contain dynamic count of channels.

Also you can get almost correct channels count from AudioTrack instance without the need of another one variable, but when i tested it against files with more than 2 channels - i got crashes. Function findFormatFromChannels works only with two variants (1 and 2 channels), and returns incorrect -1 in other cases (which cause crash).

You can also make additional option in settings, that turns on this way of channels handling.

Also check this link, it might be helpful (not sure) http://code.google.com/p/chromium/issues/detail?id=256851

Thank you.

Harry


Ph1b:
So you would recommend not getting it from the AudioTrack?

PR #171, opened 2015-04-26
Ukrainian translation

gladk:
Please add Ukrainian translation.


Ph1b:
Thanks very much!. There are some missing:

https://www.transifex.com/projects/p/material-audiobook-player/resource/stringsxml/

Issue #194, opened 2015-05-23
Use track metadata if available


kevmitch:
In the root menu for selecting books, the header at the top of the screen when the book is selected, and on the lockscreen, it displays the filename of the containing folder instead of the contents of "Album"/"TALB" tag in mp3 files (haven't tested other formats), which is typically used to identify the book title.

Similarly, the track list when a book is selected and the on lock-screen only a number identifying the currently playing track is displayed rather than the contents of the "Title"/"TIT2" tag. Strangely, with audiobooks where "TPOS" is set to group the book into first/second half, this value is displayed in preference to "TRCK" resulting in the very bad behaviour that ~20 tracks are all titled "1" and the final ~20 are titled "2".

Finally, the "Artist" / "TPE1", typically used for the book's author don't seem to be displayed anywhere.

As reported by id3v2 -l, a typical [emusic.com](http://emusic.com) audiobook's mp3 tags look like

```
id3v1 tag info for 01.part_1_of_6.mp3:
Title  : Part 1 of 6                     Artist: Douglas Coupland              
Album  : Hey Nostradamus!                Year: 2010, Genre: Blues (0)
Comment:                                 Track: 1
id3v2 tag info for 01.part_1_of_6.mp3:
TFLT (File type): MPG/3
TIT2 (Title/songname/content description): Part 1 of 6
TPE1 (Lead performer(s)/Soloist(s)): Douglas Coupland
TPE2 (Band/orchestra/accompaniment): Douglas Coupland
COMM (Comments): ()[]: Narrated by , John Randolph Jones, Jenna Lamia, Jillian C
TALB (Album/Movie/Show title): Hey Nostradamus!
TCON (Content type): Audiobooks (255)
TYER (Year): 2010
TRCK (Track number/Position in set): 1
TSIZ (Size): 33756171
APIC (Attached picture): ()[, 3]: image/jpeg, 17737 bytes
USER (Terms of use): [Cop]: yright notice. These materials are protected by United States copyright law and by international treaties. You may use these materials for your personal, non-commercial use only. You may not otherwise reproduce, distribute, publicly perform, publicly display, or create derivative works of these materials, unless authorized by the appropriate copyright owner(s).
```

And a [downpour.com](http://downpour.com) 

```
id3v2 tag info for 1.01.dedication.mp3:
TIT2 (Title/songname/content description): Dedication
TPE1 (Lead performer(s)/Soloist(s)): By Larry Niven
[...]

PR #197, opened 2015-05-25
Chapter: fix for filenames containing "."

kevmitch:
This addresses part of #194.


Ph1b:
Thanks. Implemented this while preparing data structure for getting id3 tags.

PR #216, opened 2015-06-08
App Install Location

SweetLion:
The internal memory of my device is quite small, this PR should allow users to easily move the app to the microSD.

"To allow the system to install your application on the external storage, modify your manifest file to include the android:installLocation attribute in the element, with a value of either "preferExternal" or "auto"." (source: https://developer.android.com/guide/topics/data/install-location.html)

Thank you very much for your awesome app :)


Ph1b:
Thanks for the efford.

But the document you cited refers this:

>App Widgets
>Your App Widget will be removed from the home screen. When external storage is remounted, your App Widget will not be available for the user to select until the system resets the home application (usually not until a system reboot).

This would affect the usage of the app as nobody likes disappearing widgets.

PR #241, opened 2015-06-27
Fixed issue where snackbar's theme was not getting set properly. closes #233

### LINKED ISSUE #233: Snackbar: Not themed
The snackbar does not theme accordingly to the theme that has been set.

dark-seizo:


Ph1b:
Merged 1f08d0e28994a8c28aeeead3bd50d81767bd9cfb

Note: You can always get the context directly from by [View.getContext()](http://developer.android.com/reference/android/view/View.html#getContext())

dark-seizo:
Yeah the way I was working with the Context was a little squirrelly.  I figured there was a better way, thanks for the tip.

PR #263, opened 2015-07-20
Update Ukrainian and Russian translations

gladk:


Ph1b:
Merged into develop, thanks.

Issue #199, opened 2015-05-27
Sleep timer count down: improve visual clue when enabled


NeuroNex:
**How can we know if the sleep timer is enabled?**
In version 2.1.9.1: we can notice that the sleep timer icon is slightly different. Suggestion: display a count down timer somewhere in the screen (lower left or right, or even in overlay on top of the album cover).

Ph1b:
Approved. When I find time, I'll implement a little circle in the top right corner (like the current playing icon on the BookShelf) displaying the time left for the sleep timer. If someone wants to implement this, I'd appreciate this ;-)

Ph1b:
Thanks a lot skynetua !

PR #278, opened 2015-08-07
Sleep timer count down: improve visual clue when enabled #199

skynetua:
Works ok for me, please review.

### DIFF
--- a/audiobook/src/main/java/de/ph1b/audiobook/activity/BookPlayActivity.java
+++ b/audiobook/src/main/java/de/ph1b/audiobook/activity/BookPlayActivity.java
@@ -6,6 +6,8 @@ import android.content.Intent;
 import android.graphics.drawable.Drawable;
 import android.os.Build;
 import android.os.Bundle;
+import android.os.CountDownTimer;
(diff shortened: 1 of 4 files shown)


Ph1b:
It leaks the activity.

Ph1b:
Okay, what happens is you call showTimerCountdown in onStart. You also call it in onCreate. And you call it in onSleepStateChanged. When there is a new activity (or you rotate the device), you call that method, which creates (if the sleep timer is active) 3 instances of CountDownTimer. Each of these instances hold a reference to a TextView. This textView hols a reference to the context. This context holds a reference to the activity.
In onStop you cancel the CountdownTimer. But just the last one. So all the messages referring to the internal handler of the CountdownTimer stay. And thus the whole view hierarchy is being leaked.

skynetua:
Ok, I can remove call from onCreate - it is redundant. I thought onSleepStateChanged is called only when service set\unset timer. Why it is called while activity starting?

Ph1b:
It isnt. I merged it and applied the fix. Thanks a lot!

PR #285, opened 2015-08-21
Fix chapters comparison

igormisha:


igormisha:
I changed patch since chapter cannot be directory and therefore no need to compare as files.

PR #350, opened 2015-11-20
Playstate notify for media key events.

aperomsik:
Using Material Audiobook Player in my car, I was quite happy with the way it worked, except for one thing: when I use my car's audio controls, the audio skips forwards or backwards, but the current time shown by the car doesn't change. With this commit, it usually does.

### DIFF
--- a/audiobook/src/main/java/de/ph1b/audiobook/service/BookReaderService.java
+++ b/audiobook/src/main/java/de/ph1b/audiobook/service/BookReaderService.java
@@ -211,6 +211,7 @@ public class BookReaderService extends Service implements AudioManager.OnAudioFo
 
     private boolean handleKeyCode(int keyCode) {
         Timber.v("handling keyCode: %s", keyCode);
+        boolean handled = false;
         switch (keyCode) {
             case KeyEvent.KEYCODE_MEDIA_PLAY:
             case KeyEvent.KEYCODE_MEDIA_PAUSE:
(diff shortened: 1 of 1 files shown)


Ph1b:
Hey @aperomsik 

Thanks for the work. I don't fully understand what the issue is here.

There are two subscriptions the notifyChange is triggered from. One is from the subscription to the play state, [here](https://github.com/Ph1b/MaterialAudiobookPlayer/blob/d5e4b44577e692b31062707a058794144928f9dc/audiobook/src/main/java/de/ph1b/audiobook/service/BookReaderService.java#L176)

And the other one is when the content of the book (this is also the time as the Book class is fully immutable) is changed, see [here](https://github.com/Ph1b/MaterialAudiobookPlayer/blob/d5e4b44577e692b31062707a058794144928f9dc/audiobook/src/main/java/de/ph1b/audiobook/service/BookReaderService.java#L169).

So it would be great if you could find out what exactly causes this. On of the existing mechanisms should already catch this issue.

aperomsik:
Hi @Ph1b . The key to this issue seems to be the fact that the car's position display will only update to reflect the skip if mediaSession.setPlaybackState() is called. The first callback you mentioned would trigger it, but it is not hit when pushing the car's back button. The second place you mentioned seems to be called an awful lot, but it does not call setPlaybackState because of the ChangeType.METADATA flag. 

I was quite happy to [find](http://stackoverflow.com/a/17034414/589306) that I can debug in the emulator, simulating the car's back button with 

```
adb shell input keyevent 88
```

Of course that can't tell me what the car will say, but it at least helps me know if the necessary statement gets hit. 

(However, I also found that with breakpoints in notifyChange itself, I would soon find dozens of threads stopped in that function, and the emulator's player eventually crashed.)

Ph1b:
I don't get what your car is hooking up to. Where is it getting the time information from. I mean the play state is reported correctly. What has the play state to do with the time?

For debugging: Timber is for debugging. So you should be able to see a log message each time notifyChange is called. You can also add additional logging information for debugging.

aperomsik:
mediaSession.setPlaybackState has a position argument, which apparently is propagated to Bluetooth / AVRCP. As I said, if that is not called, the car's display doesn't reflect the skip.

Ph1b:
That's very helpful, thanks.
I added a fix here can you check if it works for you? 329dec796830d3d6cc23e54ce5a26051268a697a

Anyway thanks for the pull request I think it lead to the right trace.

aperomsik:
I'll test it some time in the next few days, but it looks like it should work. Closing this pull request meanwhile.

aperomsik:
It worked. Thanks!

PR #358, opened 2015-12-11
I'd like to be able to listen at higher speed.

EdwardBetts:


bitinerant:
This would be great.  I've read some people have worked their way up to 4X normal speed.

PaulWoitaschek:
I think only a very minority of users listen at these rates.

EdwardBetts:
You're right. Most users won't listen at 2.5 times normal speed, but there is no harm in providing the option for the minority of users that would like to.

deniss-s:
2.5 is better than 2.0, however I still consider this is not enough. Currently 2.0 is a normal speed for me, but there are some readers who read way too slow, so I guess in perspective 2.5 won't be enough as well. By the way, VLC provides playback speed up to 4.0, so I consider to use the same value.

CercaTrova:
I would also like this feature up to 4x

deniss-s:
Now I listen at 2.5 and would increase the speed even more. How about making the upper speed limit editable via config file?

CercaTrova:
Or even through settings, the user could pick the maximum/desired speed and apply it across all books like some podcast apps do.

BrandonOCasey:
I created a Fork of Voice here with a max playback speed of 6. If anyone else want to use it feel free. Also feel free to bug me to pull changes from master in the issues if I fall behind, I will try to keep this up to date though. Release will be attatched with an apk in the releases section on Github:

https://github.com/BrandonOCasey/Voice
https://github.com/BrandonOCasey/Voice/releases

CercaTrova:
This is great, already listening at 4x speed. Would you be able to add the option to apply a speed across all all books?

PR #439, opened 2016-04-04
Change max playback speed to 2.5.

skorokithakis:
I changed the max playback speed to 2.5, as 2 was a bit too slow for the max speed.


PaulWoitaschek:
#358

Issue #472, opened 2016-07-22
Playback at 3.0 speed


BrandonOCasey:
Other players such as audible support up to 3.0, I think 3.0 should be used in this player as well

PaulWoitaschek:
#255

Issue #458, opened 2016-06-08
Playback speed in excess of 2x


squashmcdermott:
Hi can we get playback speed up to 2.5x in line with other audtiobook readers please? Thanks

PaulWoitaschek:
#255

Issue #440, opened 2016-04-04
Even higher playback speed


sdakak:
Please change max playback speed to 3x. Anything less is slow.

https://github.com/PaulWoitaschek/MaterialAudiobookPlayer/blob/develop/audiobook/src/main/java/de/ph1b/audiobook/Book.kt#L54

PaulWoitaschek:
#255

sdakak:
I understand your argument about simplicity. However, please reconsider your position. I believe you are prioritizing simplicity over usability. I, and many others, regularly listen to audio books at 2.75x - 3x. I also listen to Podcasts at 3x and for this reason PocketCasts (and many other audio book apps like Smart Audiobook player) allow you to go up to 3x

Issue #255, opened 2015-07-08
Playback Speeds


ThajASara:
So far I am loving this app.  The only issue for me is the highest playback speed is 2x. I mostly listen to things faster than this (2.5-3x typically).  I would love to see this added.

dalb8:
It could also be argued that finer control is needed at the speeds < 1.0x e.g. 0.85x. Perhaps the same UI as Jump To Minute should be used.

ThajASara:
I know one really good implementation I have seen is in the AntennaPod podcatcher, which is also ope source.  They use the Presstissimo program to do it, but you can select the various playback speeds you want in your quick menu so you can choose the ones you use the most often and cycle through them quickly on the player.

intrications:
AntennaPod is discussing switching away from Prestissimo to an internal implementation.

https://github.com/AntennaPod/AntennaPod/issues/532

dalb8:
Hey, I use Browser Intercept every day! Can't believe it's not built into Android; what a great app.

MAP is using Java instead of C for the speed shifting as of the latest version.




-------- Message d'origine --------
De : Michael Basil <notifications@github.com>
Envoyé: 19 juillet 2015 15:11:26 UTC+01:00
À : Ph1b/MaterialAudiobookPlayer <MaterialAudiobookPlayer@noreply.github.com>
Cc : dalb8 <daithi_black@yahoo.ie>
Objet : Re: [MaterialAudiobookPlayer] Playback Speeds (#255)

AntennaPod is discussing switching away from Prestissimo to an internal implementation.

https://github.com/AntennaPod/AntennaPod/issues/532

---

Ph1b:
What do you mean by browser intercept?

intrications:
@dalb8 Thanks. Glad to hear it is useful.

@Ph1b Browser Intercept is just a little utility app I wrote a couple of years ago:
https://github.com/intrications/browser-intercept
https://play.google.com/store/apps/details?id=com.intrications.android.sharebrowser

Ph1b:
I changed the way it works, so you can set a more fine tuned speed now. And you set in ton the fly, so no need for clicking ok before. Just set your speed and you see if it sounds good while setting it.

About the speed:
I think the very most of the people will never us a speed above 1.5 times. Except if they think it i s funny for a few seconds. So I want to have the interface simple and don't show values most of the pople don't use. Sorry :-/

BrandonOCasey:
[...]

PR #482, opened 2016-08-23
Allow faster playback speed and more fine-grained control

BrandonOCasey:
# Information
## Why do we need to listen faster than 2x?
Audible is a comparable to this app, and it let's its user listen to books [3 times faster](http://audible.custhelp.com/app/answers/detail/a_id/4795/~/how-can-i-adjust-the-playback-speed-in-the-audible-for-android-app%3F) than they were recorded. Listening to books faster than normal is actually a very common thing to do. It allows you to finish books faster and I make heavy use of it in the Audible app. I casually listen to books on audible at 3 and would love them to go even higher (say 4-5) but it all depends on the book and the reader. This brings up a problem when I switch to using DRM free books and turn to MaterialAudiobookPlayer.

## "People won't use it"
So far there are four open issues on the subject, my own included. That means that at least 4 people would use it. I can tell you for a fact that many more would. I also don't understand how not making an option more flexible would deter people from using it. In fact I am very much deterred from using the player with the speed settings how they currently are because I listen to books so much faster. 

Here are some references with anecdotal evidence on why people listen to books at faster speeds.

* http://www.geekybloggersbookblog.com/listening-to-audiobooks-at-a-faster-speed/
* http://www.everyday-reading.com/2013/06/how-to-listen-to-audiobooks-on-double.html
* https://medium.com/ios-os-x-development/i-built-a-speed-listening-app-c6d2e3704713#.gg21707lm

# Changes
* Allow users to select playback speeds up to 6x normal
  * #472, #458, #440, #255 
* Allow users to select playback speeds in increments of 0.01
  * The current rate of 0.02 is just annoying, if we want to restrict this it should be in increments of 0.05

### DIFF
--- a/audiobook/src/main/java/de/ph1b/audiobook/Book.kt
+++ b/audiobook/src/main/java/de/ph1b/audiobook/Book.kt
@@ -34,7 +34,7 @@ data class Book(val id: Long,
 
     companion object {
         const val ID_UNKNOWN = -1L
-        const val SPEED_MAX = 2F
+        const val SPEED_MAX = 6F
         const val SPEED_MIN = 0.5F
     }
 
--- a/audiobook/src/main/java/de/ph1b/audiobook/features/settings/dialogs/PlaybackSpeedDialogFragment.kt
+++ b/audiobook/src/main/java/de/ph1b/audiobook/features/settings/dialogs/PlaybackSpeedDialogFragment.kt
@@ -29,7 +29,7 @@ class PlaybackSpeedDialogFragment : DialogFragment() {
     @Inject lateinit var db: BookChest
     @Inject lateinit var playerController: PlayerController
 
-    private val SPEED_DELTA = 0.02f
+    private val SPEED_DELTA = 0.01f
     private val MAX_STEPS = Math.round((Book.SPEED_MAX - Book.SPEED_MIN) / SPEED_DELTA)
     private val df = DecimalFormat("0.00")
 


PaulWoitaschek:
I increased it to 2.5 if its that important. Does this actually work for you? I cant understand a word if it goes beyond that value. Also I assume its bad for the calm of your mind ;-)

BrandonOCasey:
Yes 3.0 and above does work for me, it takes getting used to by gradually increasing the speed. what does the calm of my mind have to do with anything?

BrandonOCasey:
I created a Fork of Voice here with a max playback speed of 6. If anyone else want to use it feel free. Also feel free to bug me to pull changes from master in the issues if I fall behind, I will try to keep this up to date though. Release will be attatched with an apk in the releases section on Github:

https://github.com/BrandonOCasey/Voice
https://github.com/BrandonOCasey/Voice/releasesv

ArseniyShestakov:
@PaulWoitaschek would you accept PR implementing configuration of max possible speed and increments through settings screen? Then everyone could still see their x2.5 limit, but few of us who crazy enough might benefit from higher speeds. I just absolutely love your project and I wish I wouldn't need to make custom builds just for the sake of higher maximal speed.

I myself didn't understand how one could listen on higher speed up to until recently since I always supposed I would either lose some details or miss narrator performance, emotions, etc. Yet I was told on reddit that brain adjust to higher speeds once you tried it for extended period of time. And in fact brain does adjust. I just finished last few Malazan books at x2.5 and I enjoyed it as much as I did when I listened at just x1.

So while I unsure how much faster I could personally go and it's obviously depend on narrator I sure that at least for some books speed like x3 and faster could work just fine.

satvikpendem:
@BrandonOCasey there's a new update, would you mind updating your fork?

satvikpendem:
@PaulWoitaschek would you add an option like @ArseniyShestakov was saying about changing speed manually within the settings, as well as setting a global playback speed instead of changing them for individual books?

ArseniyShestakov:
If I ever become a bit less lazy I could make appropriate PR or might be fork if required.
For now I just stuck using 2.5x and can playback some of my content at higher speed in other apps.

satvikpendem:
@PaulWoitaschek Please, please at least increase the speed limit to 4x, or allow us to set up our own min/max speeds. This is the **only** feature keeping me on the 3.6 version, and I'd like to try out the 4.0 version but I just can't. That's how important this feature is to me, and I'm sure many others who listen to books at high speeds. I would make a PR myself but I don't know where or how to change this value, and maintain the fork.

simonhull:
I'm going to add my voice to this too and request faster playback speeds. Maxing out at 4x would be lovely.

BrandonOCasey:
I updated my fork https://github.com/BrandonOCasey/Voice/releases

satvikpendem:
Thanks so much! Would you mind posting build instructions and how to perform this speed change for future versions as well?

Issue #301, opened 2015-09-11
Add an option to directly pick covers from the file system


PaulWoitaschek:
Rename 'edit cover' to 'Cover from internet'
Add a new option to select a cover from file system.
This shows a file picker which lets you select image files from disc. After a selection was made, open the drop dialog and set the cover.

PaulWoitaschek:
Many thanks to @alilotfi !

If you want to implement other features let me know 👍

alilotfi:
Thank you very much.
I love Kotlin and your project is an awesome playground with interesting challenges, so I like to participate in implementing other features too.
I will be happy if there is anything you would remind me.

PR #491, opened 2016-09-23
Add capability to add cover image from file systerm

alilotfi:
This commit:
- [X] Renames "Cover" to "Cover from internet"
- [x] Adds "Cover from file system" option
- [X] Allows user to pick an image, then crop it and assign it to Audiobook

(According to approved feature #301)


PaulWoitaschek:
Thanks very much! Unfortunately there were some breaking changes in the mean time.

This could be implemented from the `BookShelfController` class. Controllers also have an `onActivityResult`.
* If you change language keys please change the english translations only. The others translations will be synced through Transifex.
* The Crop intent is not reliable to be available on each device. Also it can not guarantee to be square.
* In previous versions of this app there was a dialog which was able to handle cover cropping, this could be used. -> [Latest version](https://github.com/PaulWoitaschek/MaterialAudiobookPlayer/blob/1a4dc61f7ef75aa6579fff8567653a6768208778/audiobook/src/main/kotlin/de/ph1b/audiobook/dialog/EditCoverDialogFragment.kt)

alilotfi:
Thank you very much for your tips.
About `BookShelfController`, do you mean `BookShelfPresenter`? Or I have to write the controller myself and delegate `Activity`'s `onActivityResult` to it?

PaulWoitaschek:
I mean [BookShelfController.kt](https://github.com/PaulWoitaschek/MaterialAudiobookPlayer/blob/develop/audiobook/src/main/java/de/ph1b/audiobook/features/book_overview/BookShelfController.kt)

alilotfi:
Because calling `show()` will raise:
`java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState`
The reason is that, when we show the menu dialog to user, activity saves it's instance before being paused, so before resuming again, we cannot attach a fragment to it.

PaulWoitaschek:
Thanks 👍

PR #509, opened 2016-11-22
File scanning, gestures, google drive

TimurShakirov:


PaulWoitaschek:
Please describe what you want to do. Please use the same code style. Please follow general naming practices like only capitalising variables. Please check out the latest source before making a pull request. Please do not leave dead or out commented code in the source code.

PR #512, opened 2016-11-26
Control Media Broadcasts

HumbleBeeBumbleBee:
Control Media Broadcasts with MaterialAudiobookPlayer or Simple Last.fm
Scrobbler.


PaulWoitaschek:
Can you describe why you added a setting for that?
If I understand correctly the only missing part was the package name?

HumbleBeeBumbleBee:
A user requested this feature. I added both so the user would have redundancy. This allows you to choose either method. It is correct however that for SLS only package name is required.

HumbleBeeBumbleBee:
Other scrobblers may not have this functionality.

PaulWoitaschek:
Which functionality?

HumbleBeeBumbleBee:
Disabling applications. An example is the official Last.fm scrobbler.

PaulWoitaschek:
Thanks for the pull request. I prefer not to have an additional setting.

PR #532, opened 2017-01-10
Updated el (Greek) translation

yiannakis:


PaulWoitaschek:
Thanks for the translations!

Translations are managed through [Transifex](https://www.transifex.com/material-audiobook-player/material-audiobook-player/) and imported from there. It would be great if you could change them there.

PR #570, opened 2017-03-30
Use title from meta data for book name if album is undefined

mttronc:
This PR makes the media scanner use `METADATA_KEY_TITLE` for `bookName` if `METADATA_KEY_ALBUM` is undefined.

Use case: You have an audio book mp3 file with only Title and Artist ID3 tags set.
MaterialAudiobookPlayer expects the album tag to contain the book's title and therefor ignore the title tag.
This PR fixes this behaviour.


PaulWoitaschek:
So it ignores the file completely?

mttronc:
If the album tag is empty it uses the title tag by setting `bookName = chapterName`

If neither album nor title tag are set, it will fall back to the file name because of
`chapterName = if (fileName.isEmpty()) input.name else fileName`

Hope that answeres your question.

PaulWoitaschek:
Thanks for the pull request. I implemented it at another place.

Issue #486, opened 2016-09-14
Lock screen images are ugly


larsey:
Can we please have a setting that turns off that feature? So the app does not replace the lock screen background?

PaulWoitaschek:
What's ugly about them?

There won't be a setting for such a minor thing.

larsey:
They are ugly because not all audio books comes with a image, that is used to set the lock screen image, that is high quality and made for that purpose.

I have a couple of audio books and it seems to be a 100x100 thumbnail stretched out to my phone 1920 x 1080 resolution. Hopefully you may think how bad it looks!

It should not be a big deal to implement? Just add a ui switch that controls wherever it should set the lock screen. Surly it cant be more than a couple lines of code since you already are doing this.

PaulWoitaschek:
My first design principle is minimalism. So no setting / UI component that I don't think is absolutely necessary will reach the app.

But it's a good point. I'll check the quality of an image and simply set the default cover if it's too bad.

nioncode:
I would also prefer to completely disable cover art on the lock screen.

MrReSc:
I completly agree with larsey. The thumbnail on a HiDPI screen looking very ugly.
You should simply disable the cover art on the lockscreen.

sandybeachit:
Is this a duplicate of # 581 and if so can we close it off ?

gwarser:
Form me it is a problem with lock screen control buttons - they are barely visible on some cover art images.

PaulWoitaschek:
Can you upload a screenshot where it's barely visible?

sandybeachit:
It could be the cover art is making navigation difficult? But without a 
screenshot I'm just guessing, could you upload a screenshot please.

MrReSc:
For me, it's just an optical thing. I nerver had an issue with the control buttons. It just looks ugly in case of low resolution cover art. I would prefer a Preference for toggling cover art on lockscreen.

gwarser:
> Can you upload a screenshot where it's barely visible?
I cannot take screenshot of lock screen. 
Anyway, probably this cannot be seen on screenshot - it's subjective, not only about color, but also about stripes and edges, and bright sun :)
I may try to take photo.

gwarser:
Photos taken i dark room, so they are not 100% representative
Even default MAP background can be problematic:
[...]

PR #581, opened 2017-05-04
Preference for toggling cover art on lockscreen

nioncode:
As discussed in #486, some people prefer to not get their wallpaper replaced by an audiobook's cover.

For example, I have carefully selected a couple of my photos that look great as wallpapers.
However, I rarely get to see them, because I often have MaterialAudiobookPlayer running.
99% of the cover art of my audiobooks is simply boring, though. Most of the time, it's just the book title printed in a fancy way, which does not look good at all.

### DIFF
--- a/audiobook/src/main/java/de/ph1b/audiobook/features/settings/SettingsController.kt
+++ b/audiobook/src/main/java/de/ph1b/audiobook/features/settings/SettingsController.kt
@@ -58,6 +58,14 @@ class SettingsController : BaseController() {
         .bindToLifeCycle()
         .subscribe { themeDescription.setText(it.nameId) }
 
+    // show cover art on lockscreen
+    setupSwitchSetting(
+        id = R.id.showCoverArtOnLockscreen,
+        titleRes = R.string.pref_show_cover_art_on_lockscreen,
+        contentRes = R.string.pref_show_cover_art_on_lockscreen_hint,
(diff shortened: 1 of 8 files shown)


PaulWoitaschek:
My intent was to check for the quality of the artwork, not to add an additional preference.

nioncode:
Currently, there are two problems:
1. Bad quality artwork: this looks ugly, no matter what your taste is. That's why you proposed to check for the quality and show a default cover or something if the quality is not good enough.

2. Simply ugly artwork: this has a good enough resolution, but the picture itself is simply not nice to look at, especially compared to beautiful wallpapers that are chosen explicitly for that purpose. I (and others in the linked issue thread) do not want to see their wallpaper replaced at all. That's what this PR tackles.

Could you please reconsider merging this? I'd like to avoid managing my own fork just for this single feature.

PaulWoitaschek:
In the case that the cover does not look good I suggest finding a better one. You can use the cover picker and use any image from the internet you like.

As much as I appreciate pull requests I think this is a rather edge case I don't want to confuse anyone with.

ypid:
I would find it useful. Thanks to @nioncode for the PR :+1:

Maybe a expert/advanced settings toggle preference could be used to hide such settings which might "confuse" people. This is done in many applications/appliances/whatever to make all of the users happy :wink: @PaulWoitaschek Would that be an option?

PR #592, opened 2017-05-23
fix build issues from command line

Eifelschaf:
This fixes two issues when building form command line. FOr some reason the "all" argument made the ndk compiler crash, but just putting all supported architectures in fixed that.


Eifelschaf:
OK. I guess 60 seconds may mot be enough to build everything on the remote server?

PaulWoitaschek:
Feel free to remove the timeout parameter at all for all calls. I guess CI will time out itself.

Eifelschaf:
looks like the timeout is needed after all (or the flac compilation fails), but now it looks like some test is stuck.

PaulWoitaschek:
You can remove the timeout but leave the waitFor

Eifelschaf:
I see my fault :) I'll update the code again.

PaulWoitaschek:
Thanks :)

PR #606, opened 2017-06-10
add a margin to ease cropping exactly (both to the webview and imageview for local files)

### LINKED ISSUE #540: Margin when cropping cover from the internet
When you want to illustrate a book by downloading a cover from the internet, the crop tool has a margin of something like 5px at the left and right of the screen that can not be included the crop. I find this fairly annoying.

Eifelschaf:
This should fix #540 

The margin is added before the crop and removed if canceled


exoplayer-flac/src/main/.gitiginore
```diff
@@ -1 +0,0 @@
-libs/
```
Eifelschaf: The file was named incorrectly and the correct file is right beside it. It had an 'i' too much

exoplayer-flac/src/main/jni/flac_parser.cc
```diff
@@ -22,6 +22,7 @@
 
 #include <cassert>
 #include <cstdlib>
+#include <cstring>
```
Eifelschaf: For some reason the compiler complained about a missing string import. It was a quick fix from Android studio.

PaulWoitaschek:
It does not really fix the issue. Now the selection is inaccurate. When you match the selection exactly with the borders of an image, you have some thick padding on the new screen.

Eifelschaf:
I'll check it with the new screen

PR #608, opened 2017-06-14
Proofreader Version

JimmXinu:
(In light of both your minimalist feature philosophy and the rather dodgy way I implemented this, I doubt that you'll want to include this PR, but I thought I should at least offer it.)

I've added some features to MaterialAudiobookPlayer to aid me in proofreading text.

My use case is: 
1. I text-to-speech a text.
1. I listen to it while exercising, noting each mistake in the text that I find.
1. I later review these mistakes and make corrections to the original next.

To aid with that, I've modified MaterialAudiobookPlayer to add these features:

- When I double-click play/pause on my headphones while playing: the player skips backwards a configurable number of seconds (even across file borders) and adds a 'Quick Bookmark'.  This generally puts the bookmark before the error and gives me an audible confirmation that it was recorded.

- When I tap a 'Quick Bookmark' in the bookmark list, it immediately plays from that point without closing the bookmark list and toggles the bookmark text between 'Quick Bookmark' and 'X Quick Bookmark' to indicate which marks I've already reviewed and reduce navigation taps while doing so.

- The popup menu on each bookmark has a 'Delete All Bookmarks' option.  I frequently generate 30+ corrections an hour.  A way to delete them all at once greatly speeds clean up after review.  I acknowledge this is not a great place for it.

Thank you for making MaterialAudiobookPlayer available.  I was able to have a player that did exactly what I want in a matter of a few days starting from your code.

### DIFF
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -11,6 +11,8 @@
             <option value="$PROJECT_DIR$/audiobook" />
             <option value="$PROJECT_DIR$/buildSrc" />
             <option value="$PROJECT_DIR$/core" />
+            <option value="$PROJECT_DIR$/exoplayer-flac" />
+            <option value="$PROJECT_DIR$/exoplayer-opus" />
           </set>
         </option>
         <option name="resolveModulePerSourceSet" value="false" />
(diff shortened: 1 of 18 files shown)


PaulWoitaschek:
Thanks a lot, this looks interesting.! As you have guessed I think this is not a feature made for most of the people, so I'm closing this PR.

However feel free to maintain your fork and keep it updated. The nice thing about open source is that you can do stuff like this :)

PR #614, opened 2017-06-20
Move bookmark menu icon to left side so it's never hidden by '+' button.

JimmXinu:
If you look at the bottom of this capture, you'll see that the bookmark menu button for the last entry is obscured by the circle '+' button.

This change switches the bookmark menu buttons to the left side to avoid that.

addbuttonhidesmenubutton


PaulWoitaschek:
Thanks for the PR! I really want to keep the controls to the right side. We had the exact same problem before with the book overview list screen.
I added the same solution here (bottom padding).

PaulWoitaschek:
Btw, if you want to create a bookmark, you don't have to set a title. You can just press "SET", or even the checkmark of the keyboard and it will use the title of the current track.

PR #620, opened 2017-06-25
show covers when listing books on Android wear

Eifelschaf:
I saw that when listing the books on Android wear the cover images are missing. It feels like this is not an ideal solution for the tiny thumbnails (perhaps we should scale the image in the call), but it looks very nice.

@PaulWoitaschek What's your opinion on this?

### DIFF
--- a/audiobook/src/main/AndroidManifest.xml
+++ b/audiobook/src/main/AndroidManifest.xml
@@ -105,6 +105,15 @@
           android:name="android.appwidget.provider"
           android:resource="@xml/widget_info"/>
     </receiver>
+      <provider
+          android:name="android.support.v4.content.FileProvider"
+          android:authorities="de.ph1b.audiobook.coverprovider"
+          android:exported="false"
+          android:grantUriPermissions="true">
+          <meta-data
+              android:name="android.support.FILE_PROVIDER_PATHS"
+              android:resource="@xml/cover_paths" />
+      </provider>
   </application>
 
 </manifest>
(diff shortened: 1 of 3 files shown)


PaulWoitaschek:
I don't like the idea of creating bitmaps for each single book. Did you check if we can use a file provider and just refer to the already existent cover files?

Eifelschaf:
we can use a provider, but I guess it won't work with the file provider default implementation. We can set a URI, but this URI has to be resolved somehow and to be honest I could not find where to put the provider. If you can help me with a location (or if there perhaps already is one that I could extend) I should be able to implement it.

I already tried to extend the `BookUriConverter` but I wasn't sure where the URIs would be requested.

PaulWoitaschek:
https://developer.android.com/reference/android/support/v4/content/FileProvider.html

PaulWoitaschek:
Once set up, the topic:

Generating the Content URI for a File

Explains how to get an uri. I have done this several times, so let me know if you need any help.

Eifelschaf:
I'll see what I can do.

Eifelschaf:
Here we go. Now the Android wear app has access to our covers via URIs. THanks for the tip with hte file provider!

Eifelschaf:
@PaulWoitaschek I have several changes for Android auto (the skip/previous actions) but they also need the changes in here to display cover art. Do you mind if I change the pull request and add everything in here and change the description?
Like this the wear specific changes would actually be rather small compared to the Android auto ones.

PaulWoitaschek:
You can just make a branch based on this one and open a new pr from the other branch. Currently I'm moving so I can't review it.

PaulWoitaschek:
Thanks :)

PaulWoitaschek:
Please make sure to checkout the latest changes (I restructured the class a little).

PR #621, opened 2017-06-29
Enhancements for Android Auto

Eifelschaf:
I added skip to next/previous actions to the Android auto screen (which required me to take away the default fast forward and rewind actions and replace them with custom actions doing the same and using the proper icons)

The default screen will still display the rewind/fast forward actions and the additional ones are hidden behind an overflow menu or in another secondary space (this seems to depend on the device it's shown on).

I also managed to get the current book displayed immediately if there is one.

### DIFF
--- a/audiobook/src/main/java/de/ph1b/audiobook/playback/PlaybackService.kt
+++ b/audiobook/src/main/java/de/ph1b/audiobook/playback/PlaybackService.kt
@@ -68,6 +68,7 @@ class PlaybackService : MediaBrowserServiceCompat() {
   @Inject lateinit var bookSearchHandler: BookSearchHandler
   private lateinit var mediaSession: MediaSessionCompat
   private lateinit var changeNotifier: ChangeNotifier
+  private var carConnected: Boolean = false
 
   private val audioFocusListener = AudioManager.OnAudioFocusChangeListener { focusChange ->
     i { "audio focus listener got focus $focusChange" }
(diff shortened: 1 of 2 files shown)


PaulWoitaschek:
The default rewind instead is skip is really important because of several things:

* On pre lollipop there is only next previous on the lockscreen
*  Headset controls send next/previous events

Isn't it possible to leave the default events?

Eifelschaf:
the problem is that Android gives the skip and previous actions priority in the UI and you wanted to keep the rewind/fast forward as the primary actions. If we keep them in Android Auto will show these actions on the main screen.

Does lollipop not support the custom actions? I saw no API limitation on that field, but I'll see in a simulator what happens.

PaulWoitaschek:
The issue is not about lollipop. It's about that instead of our notification, pre lollipop a screen with next/previous is shown which is not customizable.

Eifelschaf:
Okay... Then we might not be able to work like that and need to keep the current skip/previous hack. I will undo this part of the change and see if I can get the chapter listing inside a book done (I've seen it in some players, but for now I'm not sure how it's done.

Eifelschaf:
I might have a fix to keep everyone happy: I do amn API check. Android auto requires lollipop or above and our issu only exists below. I will add those actions if we are running below lollipop and keep the handline with rewind/fast forward in that case. All the others should be able to see the custom actions.

I will see what the headset actions do when I find the time, but right now my headset won't send any actions to my phone.

PaulWoitaschek:
The issue existed above too. When you long press the louder / quieter button of a headset you receive a next / previous event.

But we can checked if Android auto is connected and act based on that.

Eifelschaf:
We could do that indeed.

Paul Woitaschek <notifications@github.com> schrieb am Do., 29. Juni 2017,
20:38:

Eifelschaf:
Headset controls are definitely broken by that. Thanks for the heads up.

Matthias Kutscheid <eifelschaf@gmail.com> schrieb am Do., 29. Juni 2017,
20:39:

Eifelschaf:
Do you think we could store teh conneciton state of the car globally somewhere instead of this variable I added?

PaulWoitaschek:
> Do you think we could store the connection state of the car globally somewhere instead of this variable I added?

I refactored the receiver logic to a separate class. I'll make some further refactoring for the dependency wire-up with the next commits.

PR #662, opened 2017-10-15
Implementing playlists in MediaSession

Eifelschaf:
These commits contain an implementation of playlists in the mediasession. The playlist for each book consists of it's chapters and will only be displayed if there is more than one chapter. 

Additionally I fixed one of the test classes.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -24,7 +24,7 @@ android {
   buildToolsVersion versions.buildTools
 
   defaultConfig {
-    applicationId "de.ph1b.audiobook.matthias"
+    applicationId "de.ph1b.audiobook"
     minSdkVersion versions.minSdk
     targetSdkVersion versions.targetSdk
     versionCode 210
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -24,7 +24,7 @@ android {
   buildToolsVersion versions.buildTools
 
   defaultConfig {
-    applicationId "de.ph1b.audiobook"
+    applicationId "de.ph1b.audiobook.matthias"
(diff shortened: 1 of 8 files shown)


PaulWoitaschek:
So did that actually work for you? If I understand correctly, the `CHAPTER_ID` was never parsed so I think I fixed it in 
d85aa6b495abec3e11b150260f2ec1ab4978e33e

Also in this block:

```kotlin
      BookUriConverter.CHAPTER_ID -> {
        val id = bookUriConverter.extractBook(uri)
        currentBookIdPref.value = id
        onPlay()
      }
```
shouldn't the chapter id also be parsed so this can work correctly?

Eifelschaf:
Actually this is just to have unique IDs. What's used for playback is the
numeric item id in the playlist

Paul Woitaschek <notifications@github.com> schrieb am So., 15. Okt. 2017,
21:51:

Eifelschaf:
Works really well by the way.

Matthias Kutscheid <eifelschaf@gmail.com> schrieb am So., 15. Okt. 2017,
21:55:

PaulWoitaschek:
Hm, for me not. It does not show the book contents but only the chapters after the current chapter. So there is no way for navigating tracks back from android auto.

Eifelschaf:
Yeah, that's an auto thing. And this is also why I left the skip back
action in. On wear we see the whole list. Do you test with the simulated
head unit or phone app? Because in the phone app I can't see the extra
actions.

Paul Woitaschek <notifications@github.com> schrieb am So., 15. Okt. 2017,
22:53:

PR #664, opened 2017-10-18
fix flickering of the book listing if a queue is used

Eifelschaf:
Flickering is fixed. Please include this into your release to make the queue work well.


app/src/main/java/de/ph1b/audiobook/playback/utils/ChangeNotifier.kt
```diff
@@ -92,10 +92,11 @@ class ChangeNotifier @Inject constructor(
         .setActiveQueueItemId(book.chapters.indexOf(book.currentChapter()).toLong())
         .build()
 
-    appendQueue(book)
+//    appendQueue(book)
```
PaulWoitaschek: Please remove commented code.

Eifelschaf:
Done

PaulWoitaschek:
Thanks :)

PR #669, opened 2017-10-23
add more tests to book, convert some funs to vals

jozic:


PaulWoitaschek:
Thanks!

Btw, truth has these handy subjects which can make tests really fluent.

See e1f7bee589333710b0d7a2113b5a1b37047730be

jozic:
That's neat, have you  looked at kotlintest? https://github.com/kotlintest/kotlintest

I use scalatest (which is same thing but for scala)  daily at work and it helps to make our tests readable and fluent.
I can try it on Voice if you don't mind

PaulWoitaschek:
I also had a look at `Spek` but for now I kind of prefer to have regular junit tests, especially with the neat features juni5 will bring. (or brings already)

Issue #616, opened 2017-06-21
Feature Request: Autoplay when Bluetooth device connects


andremina:
Please include the following details in your bug report:

App-Version:
Phone Name: Pixel XL
Android Version: 7.1.2

andremina:
Best audiobook player on Android right now! 

I would love to see autoplay when phone connect to Bluetooth device. If it's an unnecessary hassle, would you be able to post the intents? So I can automate this with Tasker. Thanks!

PaulWoitaschek:
What intents do you mean?

andremina:
Thank you for taking the time. 

It would be what's in Step 5 here:

https://www.google.ca/amp/s/androidtrix.wordpress.com/2013/11/05/tasker-spotify-plug-load-and-play/amp/

PaulWoitaschek:
We use the default intents:

`Intent.ACTION_MEDIA_BUTTON` with `KeyEvent.KEYCODE_MEDIA_PLAY` as extra

andremina:
I tried the intent with the extra but I couldn't get it to work. 

de.ph1b.audiobook.playback.PlaybackService
How do I start that? Is it a MIME type or DATA or PACKAGE or CLASS? If it's something else what is it and should it come before the intent in the sequence in Tasker?

PaulWoitaschek:
You need to send an action of `Intent.ACTION_MEDIA_BUTTON` with the `KeyEvent.KEYCODE_MEDIA_PLAY` extra set to true to the service `de.ph1b.audiobook.playback.PlaybackService` in `de.ph1b.audiobook`.

ekorn:
Hi I have the idea to use nfc tags to play a audiobook. 
Is there also a option to choose the audiobook in the intend ?

jessecarroll:
@andremina did you ever get this working? I'm also struggling with activating the playCurrent intent from tasker. I've tried many many variations in the settings for the action but haven't been able to trigger the intent yet. Based on what Paul said, this is my best guess so far but I can't tell what is wrong with it:

![screenshot_20171011-111718](https://user-images.githubusercontent.com/32709564/31453109-9a2d61aa-ae76-11e7-91d3-1ca529dc6c59.jpg)

andremina:
I used an app called Bluetooth connect and play, from the App store

holocronweaver:
[...]

PR #710, opened 2018-03-30
Add support for resume on Bluetooth connect.

holocronweaver:
This fulfills #616. 

Note that I based this on 3.6.3 because master was crashing on my phone.

Plus I prefer the original large icon layout to the list layout, much easier to push.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -8,6 +8,7 @@
     <uses-permission android:name="android.permission.WAKE_LOCK"/>
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
+    <uses-permission android:name="android.permission.BLUETOOTH"/>
 
     <application
         android:name=".injection.App"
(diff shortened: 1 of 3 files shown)


holocronweaver:
@PaulWoitaschek Knock knock. Pull request here. =)

PaulWoitaschek:
Sorry, quite busy right now. I really appreciate the PR. However I don't think that this is a good default behavior and that pepole expect this to happen.

I also think it is so specific that it does not make sense to make it a setting.
However I would be open for a specific `Intent` for a play command.

holocronweaver:
> I don't think that this is a good default behavior and that people expect this to happen.

Disabling it by default is probably a good idea. 

I think 'start playback upon headset connection' makes sense both for wireless and wired headsets, but I can also see an argument for making them separate options. I'm fine either way.

> I also think it is so specific that it does not make sense to make it a setting.

I am not sure what you mean. If it's not always on, doesn't it need to be an option? What is the alternative?

> However I would be open for a specific Intent for a play command.

1. How would this differ behaviorally from my implementation? Would you have to push play after connecting the bluetooth headset? This already works as is, so there's nothing to implement for that.

To clarify, the goal is to have playback autostart on connection, like it does for wired headsets and other audio apps that support Bluetooth play on connect (this is the only audio playback app I use that doesn't support this!). No additional buttons should need to be pressed - turn on your headset and go. It's a big time saver when you connect multiple times a day like I do. It's tedious enough having to wait and hold the bluetooth headset's power on button, much less having to press another button afterwards.

2. Would it require much additional work? If it would, seeing as we both have limited time for this project, why not merge my PR as a v0 feature (after I make a few adjustments per your suggestions). Later if you or someone else wishes and has the time, come back and rewrite it as a v1 however you like. My PR as-is implements the feature I and others desire - I think disabling it by default and leaving it as an option is a fine v0. Not perfect, but at least the feature is there for those who want it, rather than waiting months or years for a v1 implementation that may never arrive.

BTW, I've been using my implementation daily for a couple months, works great! Haven't experienced any bugs.

PR #711, opened 2018-03-30
Continue playing when editing current book.

### LINKED ISSUE #706: Editing while playing
When editing while playing the book you are playing the player no longer responds.

holocronweaver:
* Setting player.playWhenReady = false when editing current playing
  book causes audio to stop and the player shows a pause button that
  does not change state when pushed.
* Besides breaking playback controls, I see no reason to stop playback when a book is edited.
* Fixes #706.

### DIFF
--- a/app/src/main/java/de/ph1b/audiobook/playback/MediaPlayer.kt
+++ b/app/src/main/java/de/ph1b/audiobook/playback/MediaPlayer.kt
@@ -135,7 +135,6 @@ constructor(
     if (player.playbackState == Player.STATE_IDLE || bookSubject.value != book) {
       Timber.i("init called with ${book.name}")
       bookSubject.onNext(book)
-      player.playWhenReady = false
       player.prepare(dataSourceConverter.toMediaSource(book))
       player.seekTo(book.currentChapterIndex, book.positionInChapter.toLong())
       player.setPlaybackSpeed(book.playbackSpeed)


PaulWoitaschek:
This does not really fix the issue. The issue itself is present because upon name editing a new book is being initialized.
But the player contents itself did not change. Pausing the player when a new book is initialized is necessary because we want to pause when the user changes the book.

holocronweaver:
So the solution is not to initialize a new book upon edit? It seems initializing books and updating the UI are intimately tied together, wasn't sure the rationale, so I went for the easy fix.

I actually prefer continuing play while changing books. Almost every audio player I use behaves this way. Why pause?

PaulWoitaschek:
The solution is to decouple the presentation part of the book from the core media part and pass that to the media player. Then reactively only reinitialize the player only when the actual chapters have changed.
I'm working on that currently. 

Just omitting the init call would lead to the player overwriting name changes.

holocronweaver:
Ahh, gotchya! Thanks for the feedback, glad you're working on it.

PaulWoitaschek:
Still has some quirks but this should do it:
 d18421131a5f013a90feb50a597fe83ce35efe63

I generally don't like the double-coupling between player that updates the repo and repo that updates the player. But I don't have a better solution currently.

PR #756, opened 2018-09-22
updated CoverReplacement to be executed on UI thread

rinormaloku:
Hi,

I tested out the current state and I was crashing when scrolling through a long list of books. The crash was due to the line below: 

`
val coverReplacement = CoverReplacement(bookName, context)
`

with the following message:

` only the original thread that created a view hierarchy can touch its views. ` 

**Important!** I am not an Android developer the solution is an outcome of some googling and checking out implementations in Voice. If this is not correct I would be happy if you give me hints and some links so that I can improve it, and learn :)

### DIFF
--- a/app/src/main/java/de/ph1b/audiobook/features/bookOverview/list/LoadBookCover.kt
+++ b/app/src/main/java/de/ph1b/audiobook/features/bookOverview/list/LoadBookCover.kt
@@ -16,6 +16,7 @@ import kotlinx.coroutines.experimental.Dispatchers
 import kotlinx.coroutines.experimental.GlobalScope
 import kotlinx.coroutines.experimental.Job
 import kotlinx.coroutines.experimental.android.Main
+import kotlinx.coroutines.experimental.android.UI
 import kotlinx.coroutines.experimental.isActive
 import kotlinx.coroutines.experimental.launch
 import kotlinx.coroutines.experimental.withContext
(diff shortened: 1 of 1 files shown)


PaulWoitaschek:
Can you post the complete stacktrace?

rinormaloku:
The stack trace:

```
09-22 15:28:02.006 6044-6782/de.ph1b.audiobook E/AndroidRuntime: FATAL EXCEPTION: pool-3-thread-21
    Process: de.ph1b.audiobook, PID: 6044
    android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
        at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:6664)
        at android.view.ViewRootImpl.invalidateChildInParent(ViewRootImpl.java:960)
        at android.view.ViewGroup.invalidateChild(ViewGroup.java:5083)
        at android.view.View.invalidateInternal(View.java:12741)
        at android.view.View.invalidate(View.java:12705)
        at android.view.View.invalidate(View.java:12689)
        at de.ph1b.audiobook.uitools.SquareProgressView$$special$$inlined$observable$1.afterChange(Delegates.kt:74)
        at kotlin.properties.ObservableProperty.setValue(ObservableProperty.kt:41)
        at de.ph1b.audiobook.uitools.SquareProgressView.setColor(SquareProgressView.kt)
        at de.ph1b.audiobook.features.bookOverview.list.LoadBookCover$load$1.doResume(LoadBookCover.kt:58)
        at kotlin.coroutines.experimental.jvm.internal.CoroutineImpl.resume(CoroutineImpl.kt:42)
        at kotlinx.coroutines.experimental.DispatchedTask$DefaultImpls.run(Dispatched.kt:149)
        at kotlinx.coroutines.experimental.AbstractContinuation.run(AbstractContinuation.kt:19)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
        at java.lang.Thread.run(Thread.java:818)

```

Though I saw some other exceptions aswell but this was the only that broke the app and made subsequent starts crash immediately

PaulWoitaschek:
Thanks. The line you posted is not the issue.

The issue is that the color was changed on the IO dispatcher.
On Android we don't want to block the main thread as it would cause the UI to lag.
Therefore the `file.exists()`, `file.length()` etc are calls made on the IO thread.

However we are not allowed to change the UI from any other thread than the UI thread.
So it's acutally [this line](https://github.com/PaulWoitaschek/Voice/commit/d9977cb642d35dbc166e63e738a5b38b29a44861#diff-4eaca8b5ff215ed77ddcbe9624fb01d5R55) that needs to be called on the UI thread.

Thanks for contributing and reporting!

PR #764, opened 2018-10-14
Audioplayer redesign

newhinton:
I have played with some ideas, what do you think about it?


PaulWoitaschek:
Thanks for your contribution.
However to me this creates too much visual clutter. I don't want to build a recognizable app but rather keep it as plain and simple as possible.

PR #768, opened 2018-10-14
Added Recursive folderadding

newhinton:
This enables the user to add multiple folders at once, which is useful if you have a well structured folder like this:

\audiobook\author1\book1
\audiobook\author1\book2
\audiobook\author2\book1
\audiobook\author2\book2
\audiobook\author2\book3
\audiobook\author3\book1


PaulWoitaschek:
Thanks for your contribution.
However this is not how adding books work. The app rather monitors a folder (see `BookAdder.kt`.)
I generally like the idea of being flexible to this folder structure. However the way it is now is confusing enough for the user so I'd rather rethink how we communicate it better to user in general before implementing anything.

Gettyk:
That would be nice to have this recursive features. And even better: to have a full navigation of the folders (like this [app ](https://play.google.com/store/apps/details?id=mdmt.sabp.free)has). That way may be less confusing for the users

Issue #767, opened 2018-10-14
Migrate call classes extending DialogFragment to extend DialogController


PaulWoitaschek:

Plastix:
I'm working on this!

PR #769, opened 2018-10-14
Migrate DialogFragments to DialogController

Plastix:
This commit also renames any fragment classes to suffix with
'controller'.
Completes https://github.com/PaulWoitaschek/Voice/issues/767

### DIFF
--- a/app/src/main/java/de/ph1b/audiobook/features/bookOverview/EditBookTitleDialogController.kt
+++ b/app/src/main/java/de/ph1b/audiobook/features/bookOverview/EditBookTitleDialogController.kt
@@ -3,12 +3,12 @@ package de.ph1b.audiobook.features.bookOverview
 import android.app.Dialog
 import android.os.Bundle
 import android.text.InputType
-import androidx.fragment.app.DialogFragment
 import com.afollestad.materialdialogs.MaterialDialog
 import de.ph1b.audiobook.R
 import de.ph1b.audiobook.data.Book
 import de.ph1b.audiobook.data.repo.BookRepository
 import de.ph1b.audiobook.injection.App
+import de.ph1b.audiobook.misc.DialogController
 import de.ph1b.audiobook.misc.getUUID
 import de.ph1b.audiobook.misc.putUUID
 import kotlinx.coroutines.experimental.GlobalScope
@@ -18,16 +18,16 @@ import javax.inject.Inject
 /**
  * Simple dialog for changing the name of a book
  */
-class EditBookTitleDialogFragment : DialogFragment() {
+class EditBookTitleDialogController : DialogController() {
 
   @Inject
   lateinit var repo: BookRepository
 
-  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+  override fun onCreateDialog(savedViewState: Bundle?): Dialog {
     App.component.inject(this)
 
-    val presetName = arguments!!.getString(NI_PRESET_NAME)
-    val bookId = arguments!!.getUUID(NI_BOOK_ID)
+    val presetName = args.getString(NI_PRESET_NAME)
+    val bookId = args.getUUID(NI_BOOK_ID)
 
     return MaterialDialog.Builder(activity!!)
       .title(R.string.edit_book_title)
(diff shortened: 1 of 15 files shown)


app/src/main/java/de/ph1b/audiobook/features/settings/dialogs/PlaybackSpeedDialogController.kt
```diff
@@ -53,25 +57,31 @@ class PlaybackSpeedDialogFragment : DialogFragment() {
...
+      seekBar.progressChangedStream(initialNotification = true)
+        .map { Book.SPEED_MIN + it.toFloat() / FACTOR }
+        .doOnNext {
+          // update speed text
+          val text =
+            "${activity!!.getString(R.string.playback_speed)}: ${speedFormatter.format(it)}"
+          textView.text = text
+        }
+        .debounce(50, TimeUnit.MILLISECONDS, AndroidSchedulers.mainThread())
+        .subscribe { playerController.setSpeed(it) }) // update speed after debounce
```
Plastix: It is usually bad practice to subscribe without clearing the `Disposable` at some point. I thought I would add this here.

PaulWoitaschek: Let's move that logic to the `DialogController` and add an extension function `Disposable.disposeOnDestroyDialog`.

The implementation detail that's making this buggy is that `onCreateDialog` is called from the controllers `onCreateView`. But you dispose now in `onDetach`. 
When we are now in a situatio occurs where `onDetach` is called, but no `onDestroyView`, then we'll dispose the progress stream, but don't observe it when `onAttach` is called.

You can test this by opening the dialog, click on the home button and open the app again. At this point, the speed won't change upon sliding.

In general the subscription only needs to be disposed if it has a reference to the activity for example. Else it will be garbage collected along with the view. It's basically the same like with an `OnClickListener`. You don't remove all `OnClickListener`s when a view gets destroyed too.

app/src/main/java/de/ph1b/audiobook/features/bookOverview/EditBookBottomSheet.kt
```diff
@@ -58,8 +58,9 @@ class EditBookBottomSheet : BottomSheetDialogFragment() {
     dialog.setContentView(container.containerView)
 
     container.title.setOnClickListener {
-      EditBookTitleDialogFragment.newInstance(book)
```
PaulWoitaschek: Can you replace the `EditBookBottomSheet` with a `DialogController` too? It seems that `BottomSheetDialogFragment` just wraps the dialog in an additional `BottomSheetDialog` so we might just omit that.

app/src/main/java/de/ph1b/audiobook/features/bookOverview/EditBookTitleDialogController.kt
```diff
@@ -50,18 +54,14 @@ class EditBookTitleDialogFragment : DialogFragment() {
...
-    fun newInstance(book: Book): EditBookTitleDialogFragment {
-      val args = Bundle()
-      args.putString(NI_PRESET_NAME, book.name)
-      args.putUUID(NI_BOOK_ID, book.id)
-
-      val dialog = EditBookTitleDialogFragment()
-      dialog.arguments = args
-      return dialog
-    }
+    fun newInstance(book: Book): EditBookTitleDialogController =
```
PaulWoitaschek: Use a constructor

app/src/main/java/de/ph1b/audiobook/features/settings/dialogs/PlaybackSpeedDialogController.kt
Plastix: That makes sense. I'm not familiar with the `conductor` lifecycle. Thank you!

app/src/main/java/de/ph1b/audiobook/features/bookOverview/EditBookTitleDialogController.kt
Plastix: I'm sorry, but I don't know what you mean. This method is using a constructor? Are you saying we should remove the `newInstance` method?

PR #776, opened 2018-10-17
Remove custom mockito mocking function

Plastix:
The project already uses the extension library `mockito-kotlin` which
provides helper functions to make Mockito work with Kotlin. There is no
need for our custom `given` function if we can use the built in
`whenever`. This also makes it less confusing to new developers.


PaulWoitaschek:
Nice, thanks!

PR #788, opened 2018-11-09
Updated broken translation URL

adriancampos:
Since the rename to Voice, clicking the Translations option under the heart in preferences takes you to a nonexistent transifex project ([material-audiobook-player](https://www.transifex.com/projects/p/material-audiobook-player/)). Updated it to the new one.


PaulWoitaschek:
Thanks!

PR #789, opened 2018-11-10
Disable Room queries on the main thread

Plastix:
Currently our Room db builder allows queries on the main thread. This
can result in dropped frames and poor performance. With
e4e5b502f69d73c1e3b95b14db908a33c9c5e3c9, `BookStorage` no longer runs
queries on the main thread so we can disable this Room setting.


PaulWoitaschek:
The initial data fetch is still on the UI thread:
https://github.com/PaulWoitaschek/Voice/blob/master/data/src/main/java/de/ph1b/audiobook/data/repo/BookRepository.kt#L26

Did you try this out? Didn't it crash immediately?

Plastix:
Isn't it on an IO coroutine?  https://github.com/PaulWoitaschek/Voice/blob/b5cce3a55969bed9b4f945b4f0210992e033a9be/data/src/main/java/de/ph1b/audiobook/data/repo/internals/BookStorage.kt#L24-L42

I did run it and it didn't crash.

Issue #798, opened 2018-12-13
Inconsistent Data


PaulWoitaschek:
I have this crash I can't make sense of. Any idea how that could happen?

cc @Plastix 

```
Fatal Exception: io.reactivex.exceptions.OnErrorNotImplementedException: The exception was not handled due to missing onError handler in the subscribe() method call. Further reading: https://github.com/ReactiveX/RxJava/wiki/Error-Handling | java.util.NoSuchElementException: Collection contains no element matching the predicate.
       at io.reactivex.internal.functions.Functions$OnErrorMissingConsumer.accept(Functions.java:704)
       at io.reactivex.internal.functions.Functions$OnErrorMissingConsumer.accept(Functions.java:701)
       at io.reactivex.internal.observers.LambdaObserver.onError(LambdaObserver.java:77)
       at io.reactivex.internal.operators.observable.ObservableFlatMap$MergeObserver.checkTerminate(ObservableFlatMap.java:496)
       at io.reactivex.internal.operators.observable.ObservableFlatMap$MergeObserver.drainLoop(ObservableFlatMap.java:334)
       at io.reactivex.internal.operators.observable.ObservableFlatMap$MergeObserver.drain(ObservableFlatMap.java:326)
       at io.reactivex.internal.operators.observable.ObservableFlatMap$InnerObserver.onError(ObservableFlatMap.java:575)
       at io.reactivex.internal.observers.BasicFuseableObserver.onError(BasicFuseableObserver.java:100)
       at io.reactivex.internal.operators.observable.ObservableSwitchMap$SwitchMapObserver.drain(ObservableSwitchMap.java:211)
       at io.reactivex.internal.operators.observable.ObservableSwitchMap$SwitchMapObserver.onError(ObservableSwitchMap.java:140)
       at io.reactivex.internal.operators.observable.ObservableSwitchMap$SwitchMapObserver.onNext(ObservableSwitchMap.java:115)
       at io.reactivex.internal.operators.observable.ObservableMap$MapObserver.onNext(ObservableMap.java:62)
       at io.reactivex.internal.operators.observable.ObservableConcatMap$ConcatMapDelayErrorObserver.drain(ObservableConcatMap.java:470)
       at io.reactivex.internal.operators.observable.ObservableConcatMap$ConcatMapDelayErrorObserver.onSubscribe(ObservableConcatMap.java:330)
[...]

PR #800, opened 2018-12-15
[#798] Handle RxJava errors in TriggerWidgetOnChange

Plastix:
When you subscribe to an RxJava you must provide an `onError` callback
in the subscriber in case the stream errors. Otherwise, the error cannot
be delivered.

This commit adds an `onError` callback which logs to crashlytics. Note:
this does not fix the underlying issue but is more of a band-aid fix.


Plastix:
Simple fix for https://github.com/PaulWoitaschek/Voice/issues/798. I'm looking into the underlying issue. 

I haven't tested this since my Android Studio project has configuration problems at the moment.

PaulWoitaschek:
You should only handle onError if you expect an error.
Here there is an underlying bug which needs to be fixed so catching this hides the actual bug.
I expressed that further in #798

PR #801, opened 2018-12-16
[Proposal] Replace Observable<Optional> with null object pattern

Plastix:
**Background:**
Currently `BookRepository.byId(...)` returns an
`Observable<Optional<Book>>`. This means that every emission creates a
wrapper object. This commit models the no-book available case using the
Null Object Pattern so we don't create uncessary object wrappers.

**Modifications:**
* Remove our custom `Optional` implementation and the `Optional.getIfPresent()`
extension function.
* Add a static `null object` book instance in the `Book` companion
object.
* Update `BookRepository.byId(...)` to return this book object in the
error state.
* Update all callers to filter out null book objects.

_More info here:_
https://github.com/ReactiveX/RxJava/issues/4644
https://en.wikipedia.org/wiki/Null_object_pattern


PaulWoitaschek:
I'm not a fan of this. A book is a book, not a null object.

If we allow this kind of invalid books, each class consuming books would need to check  for the presence of a book.
The real issue here is that rxjava has a `Maybe` (single event which can or can't have a value), but no equivalent in the streaming world.

So kotlin and rxjava are not really so great together because if rxjava was a pure kotlin implementation, the nullity would be typed: `Observable<Book?>` in that case.

I like the optional here because it's very explicit of what that stream returns.

If you like it more kotlin idiomatic you could use a sealed class:
```kotlin
sealed class<T> Option<T>{
  data class Present<T>... :Option()
  data class Absent<T>... :Option()
}
```

Issue #834, opened 2019-01-17
[Enhancement request] Add percentage of book covered


JoelAbees:
Please include the following details in your bug report:

App-Version:
Phone Name:
Android Version:

Hi paul. First of thanks for this amazing app. While I appreciate the fact that it is a free app, I have a suggestion to improve it just a tiny bit. 

You can display percentage of the book covered so far. 

Thank you,
Joel

jaqjacek:
+1 for this sugestion

GitBugit:
I would not only prefer to have a percentage indicator but also one that shows exactly how much time I listened already and how much I have left

pedrojvaz:
Hopefully, this can be added soon

PaulWoitaschek:
I try to follow a minimal design principle and therefore don't want another element in the UI.
There is already the progress bar so you can estimate the percentage.

ronilaukkarinen:
This is a bummer. Percentage is a must-have.

PR #851, opened 2019-03-01
Added new percentage indicator

ivanmgribeiro:
This PR adds a new circular percentage indicator overlaid on the bottom right of the book cover.
Suggested in #834

### DIFF
--- a/app/src/main/java/de/ph1b/audiobook/features/bookOverview/list/GridBookOverviewComponent.kt
+++ b/app/src/main/java/de/ph1b/audiobook/features/bookOverview/list/GridBookOverviewComponent.kt
@@ -98,6 +98,7 @@ class BookOverviewHolder(
     cover.transitionName = model.transitionName
     remainingTime.text = formatTime(model.remainingTimeInMs.toLong())
     this.progress.progress = model.progress
+    this.percentageProgress.progress = model.progress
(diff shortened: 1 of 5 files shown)


GitBugit:
Could you also integrate a progress bar showing how much time is remaining etc

test2a:
@gitbugit

https://ibb.co/R6TjC9V

We already have a progress bar below each book that shows the progress. I dont know what more are you asking for ? Maybe in the book page?

GitBugit:
@test2a 🤦‍♂️ I had the wrong layout, didn't know that a different layout shows the time. But this is exactly what I meant so my bad

test2a:
@GitBugit not a problem

PaulWoitaschek:
Sorry but this won't get merged. The implementation looks really good an clean so kudos for that. However I think the new slider implemenation from the material components is good enough to pick granular values.

robertmarsal:
Hi @PaulWoitaschek this is a feature I am interested in myself and I see a couple of requests in the issue tracker for it as well so can I please check that you closed the right PR as your comment mentions picking values in a slider but this PR is about displaying the completion percentage of a book (unless I am wrong)?

Thanks

PaulWoitaschek:
Ah these were two different issues indeed. I'll answer in the linked issue.

PR #888, opened 2019-06-19
Workaround to prevent burning eyes when starting the app in the dark.

texoft:
My humble attempt to fix white screen while the app is loading. I'm using the app only in the dark and the white startup screen really hurts eyes.


PaulWoitaschek:
This is now fixed on android 10 by using the system darkmode.

SenorSmartyPants:
Doesn't help those of us on Android 9. :(

Issue #885, opened 2019-06-16
[Feature Request] Delete File


sengsational:
My personal use-case is to load podcasts into a directory on my phone from my PC (which has my podcast catcher).  

When I'm done listening, I want to delete the files as I go.

I've added a delete action (no code behind it yet).  Just wanted to see if there was a reason that, up to this point, file delete was not included.  Maybe it's a 'dumb idea'?  I like how this version puts files you've listened to already in a separate _Completed_ section, but if I'm going to "clean-up", I've had to go to a file-system app, and then I can't see which ones I've listened to already.

![DeleteAction](https://user-images.githubusercontent.com/15800701/59558201-92238080-8fba-11e9-9a49-813d518f883e.jpg)

0rAX0:
I had to delete an audiobook as well and needed to go to the file manager to look for it. I'd love to see this as well! 👍

markusd112:
I think it is a good idea to implement a delete button. But not only for the whole book (folder with several MP3's), as well for single chapters (single MP3 files)

sengsational:
Yes, I was thinking that deleting single MP3 files would be in line with deleting podcasts that have been played.

PaulWoitaschek:
Duplicate of #554

PR #901, opened 2019-08-10
File delete

sengsational:
Added delete file functionality (#885)

To test, long press a file on either of the book category or book overview and select the delete option.  The file should be removed and the list should refresh and not have the file any more.

### DIFF
--- a/app/src/main/java/de/ph1b/audiobook/features/bookCategory/BookCategoryController.kt
+++ b/app/src/main/java/de/ph1b/audiobook/features/bookCategory/BookCategoryController.kt
@@ -8,6 +8,7 @@ import androidx.recyclerview.widget.DefaultItemAnimator
 import androidx.recyclerview.widget.GridLayoutManager
 import de.ph1b.audiobook.R
 import de.ph1b.audiobook.data.Book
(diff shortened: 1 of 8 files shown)


sengsational:
This implementation needs more work; if a file is deleted while it's playing, it will continue to play and the user has lost the ability to control playback. This could be fixed by stopping playback before delete.

Pentaphon:
Any progress with this? Deleting audiobooks from the UI is a pretty useful function, especially if you listen to short books or use Voice to listen to long audio clips.

test2a:
I would suggest this option be made visible only after a book is finished so that its certain that the user has gone through the book. If a user wants to delete a book halfway though, they should have to manually reach to the end where the delete option would pop up.

Pentaphon:
> I would suggest this option be made visible only after a book is finished so that its certain that the user has gone through the book. If a user wants to delete a book halfway though, they should have to manually reach to the end where the delete option would pop up.

Or there can just be a confirmation before actually deleting the audiobook. A two-step delete.

PaulWoitaschek:
Thanks for the PR! This somehow got autoclosed when master was renamed to main. This however is now implemented.

PR #912, opened 2019-08-15
Update README.md

Dzhuneyt:


PaulWoitaschek:
Thanks, but I don't think an additional link is necessary here.

Issue #860, opened 2019-03-21
global playback speed


test2a:
a quality of life improvement but there is no global playback speed setting if you have to listen in any speed other than 1x.

PaulWoitaschek:
I understand that this is a need of some people but I also think that the vast majority people just plays at 1.0 speed so I don't want another setting.

CreamyCookie:
> vast majority people just plays at 1.0 speed

Also, without statistics I wouldn't be so sure that the vast majority listen at 1.0 speed. If 30% listen at 1.5 speed, would that be enough for you to warrant that simple addition?

> I don't want another setting.

I understand that there's some small maintenance cost associated with each new option but, in my opinion, Voice has far too few options. When I open the settings menu, there are 4 entries. 

I think these would absolutely make sense:

- default playback speed
- a slider for setting the new fade out time (I personally dislike fade out in general.)
- a toggle for whether or not you want the timer to always start (given that it's not already running) when you press play

I wouldn't even mind having to edit a text config file for these.

Anyway, thanks for creating and maintaining Voice. It's an awesome audiobook player! =)

CreamyCookie:
> vast majority people just plays at 1.0 speed

Without statistics I wouldn't be so sure that the vast majority listen at 1.0 speed. If 30% listen at 1.5 speed, would that be enough for to warrant the addition?

> I don't want another setting.

I understand that there's some small maintenance cost associated with each new option but, in my opinion, Voice has far too few options. When I open the settings menu, there are only 4 entries. 

I think these would absolutely make sense:

- default playback speed
- a toggle for whether or not you want the timer to always start (given that it's not already running) when you press play
- a slider for setting the new fade out time (I personally dislike fade out in general.)

I wouldn't even mind having to edit a text config file for these.

Anyway, thanks for creating and maintaining Voice. It's an awesome audiobook player! =)

PaulWoitaschek:
It's mostly not about the maintenance cost. I'ts about providing good defaults and having a simple ui. I'd accept a PR that reads from a settings file.
As of android Q I don't know any place for an external config file to go.

egvimo:
I would also appreciate a global setting for playback speed. Maybe it could be included in some 'Advanced Settings' like discussed in #859, or like Android is doing in the settings:

![Android-O-display-settings](https://user-images.githubusercontent.com/5247100/78470750-f838ea00-772b-11ea-8bdd-164448141c73.jpg)

[...]

PR #914, opened 2019-08-22
Add setting to allow global save of last playback speed for future playbacks

sengsational:
This PR addresses #860 (global playback speed).

Added an item to the settings page to allow the save of playback speed so that all future playback will occur at that speed.  The default setting (false) is to leave the app behavior unchanged.  If changed to "true", then if playback speed is changed during any playback, all future playback will occur at that speed.


CreamyCookie:
Does this only apply to new audiobook? I sometimes listen to multiple audiobooks, so I'd like to set a default playback speed, but that shouldn't affect existing audiobooks where I've changed the speed.

satvikpendem:
There should be an option to use global versus custom, such as a checkbox (Use Global Speed) that disables the slider, and when it's unchecked, the slider is enabled.

PaulWoitaschek:
Thank you for your implementation and taking the time contributing to voice. However for the reasons explained in #860 I don't want this to be included into Voice.

PR #929, opened 2019-11-11
Removes unnecessary comma

Potolski:
This PR removes unnecessary comma on Readme:

> Use commits with a good description, so everyone can see, what you did

to 

> Use commits with a good description, so everyone can see what you did


PaulWoitaschek:
Thanks!

PR #1007, opened 2020-07-16
Prevent continuous notification in ambient display/AOD. Fixes #983

### LINKED ISSUE #983: Display always staying awake / always-on-display gets triggered
When running Voice and turning off my screen, it immediately turns back on and it looks like the Always-On-Display is getting triggered: the lock icon gets displayed, the current time is visible as well as the battery percentage. In the middle of the screen, the Voice notification is displayed.

I have Always-On-Display DISABLED in my phone and it never shows up when using any other audio app (like PocketCasts, Google Play Music), but somehow it gets triggered by Voice.

I think this only appeared since the v5.0 update.

Phone: Google Pixel 2, Android 10

moneytoo:
This change resolves mentioned issue on my OnePlus 7 (Android 10) with enabled system option "Wake screen when you receive notifications".


PaulWoitaschek:
This was automatically closed without my notice when the master branch was renamed to main. Sorry for that! The original issue should now be fixed.

PR #1072, opened 2021-07-28
Implement Smart Recursive Folderadding

### LINKED ISSUE #809: Re-design the book adding process
Right now the whole adding process is severely flawed.

# Issues
Users...
* get confused about what the two book adding modes mean
* don't see the sd card selection in the top right
* get confused about the file selection / confirm buttons

# Current Solution
Currently there are two kind of folder structures we support:
* In a Folder each folder is recognized as a book:
  * `./book1`
  * `./book2`
  * `./book3`
* A single book
  * ./book

# Improvements

Another structure we should support is by author:
* `authorA/book1`
* `authorA/book2`
* `authorB/book1`

I think we need to re-think in a larger picture how we let users add folders.
This should be combined with an onboarding flow. Right now the user is thrown into cold water and I assume that mostly users with a technical background understand how to add folders.

I'm open for ideas! @0rAX0 cc

newhinton:
Todo: Think about Descriptions/Icons
~~Remove accidentially added gradlefile~~
~~Actually remember audiobooks that were added viar recursive adding~~

fixes #809


newhinton:
In what cases does this not work?

As soon as audiobooks are not contained in a single folder. If they are divided by CD or similar, they are not recognized as one audiobook.

How could that be fixed? 
Basically guessing. We could search the current folder for keywords in subfoldernames, and when we find them, add the current folder as a collection and stop recursing. This has the drawback that we need to provide a list with keywords, which contain all possible keywords in different languages (part, cd, teil, number etc)

Issue #1093, opened 2022-01-01
Bounty proposed: Restart sleep timer via play button on headphones (additionally to shaking)


bjohas:
It's amazing that 
* Voice has got the shake feature, to restart the sleep timer.
* Also, it's great that the restart now only happens within 30 secs of the sleep timer expiring. I.e., it's no longer possible for the book to be accidentally restarted when the device is picked up later (In Version 5.0.2, https://github.com/PaulWoitaschek/Voice/issues/954).

** Feature request ** 

I would like to be able to restart the timer via a play/pause button, typically on headphones (wired or bluetooth). For example, when listening on headphones or via a pillow-speaker, the phone may not be all that close to allow shaking to restart; however, it may be easy to restart the player via the play/pause button on headphones.

The issues https://github.com/PaulWoitaschek/Voice/issues/742headphones, https://github.com/PaulWoitaschek/Voice/issues/988 are related to this. There are two options: 
# The play/pause button has to be pressed within the 30 seconds after sleep-timer expiry.
# The sleep timer persists, i.e., play/pause button can be pressed any time after sleep-timer expiry to restart the sleep timer. However, the shake timer would be unaffected (i.e., expire 30 seconds after sleep-timer expiry).

I can see that Option 1 is consistent with the shake behaviour. However, I can see Option 2 being more useful: e.g., people forgetting to put the sleep timer on (https://github.com/PaulWoitaschek/Voice/issues/742headphones) or simply restarting listening again without having to set up the sleep timer. Therefore, I prefer Option 2.

**UI**

There would probably need to be an option in the UI somewhere. I don't really mind how that's done. Obviously, Voice takes a lean approach, so the option would need to be unobtrusive. A simple option, proposed in the other issues, is to add a check-box to the sleep timer menu, e.g. 'resume sleep timer on play' or similar.

**Bounty**

If @PaulWoitaschek agrees with the change (and would accept a PR), I'd like to offer a USD 50 bounty for the implementation of this feature (implemented as a PR on https://github.com/PaulWoitaschek/Voice as well as a downloadable apk, based off 5.0.2). 

Any interest?

bjohas:
Hi @PaulWoitaschek - any thoughts on this? Would a PR be welcome?

nydrani:
[...]

PR #1154, opened 2022-02-25
Proposing a simple fix for https://github.com/PaulWoitaschek/Voice/issues/1093

bjohas:
Hi @PaulWoitaschek 

I have **not** compiled this (as I don't have those skills), but I think https://github.com/PaulWoitaschek/Voice/issues/1093 would have a simple fix. Conceptually, the idea is
* Let sleep timer exit.
* If the player is playing at that stage, restart the sleep timer.

```
      if (playStateManager.playState == Playing) {
        Logger.i("Sleep timer finished, player is playing, restart timer.")
        start()
      }
```


PaulWoitaschek:
Thanks for the idea. Closing this until a decision on the feature was made.

PR #1341, opened 2022-07-13
Add a structure for a child ViewModel

PaulWoitaschek:
... that can provide it's items to the root bottom sheet view model.
Preparation for https://github.com/PaulWoitaschek/Voice/pull/1059


shirish87:
[A couple of suggestions](https://github.com/PaulWoitaschek/Voice/compare/multi_bottom_sheet_provider...shirish87:Voice:multi_bottom_sheet_provider?expand=1). Still, I don't think calling `onItemClicked` on every ViewModel (with an item that may not belong to it) is the way to go.

Anyway, this is [how it would work](https://github.com/shirish87/Voice/compare/multi_bottom_sheet_provider...shirish87:Voice:multi_bottom_sheet_provider_1059?expand=1) with PR #1059.

Thoughts?

PR #1342, opened 2022-07-14
BottomSheetItemViewModel.menuOrder for ViewModel submenu slots, retur…

PaulWoitaschek:
…n value for BottomSheetItemViewModel.onItemClicked

### DIFF
--- a/bookOverview/src/main/kotlin/voice/bookOverview/bottomSheet/BottomSheetViewModel.kt
+++ b/bookOverview/src/main/kotlin/voice/bookOverview/bottomSheet/BottomSheetViewModel.kt
@@ -23,17 +23,16 @@ class BottomSheetViewModel
   internal fun bookSelected(bookId: Book.Id) {
     this.bookId = bookId
     scope.launch {
-      val items = viewModels.flatMap { it.items(bookId) }
-        .toSet()
-        .sorted()
+      val items = viewModels.sortedBy { it.menuOrder }
+        .flatMap { it.items(bookId) }
       _state.value = EditBookBottomSheetState(items)
     }
   }
 
-  internal fun onItemClick(item: BottomSheetItem) {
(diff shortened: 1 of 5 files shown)


bookOverview/src/main/kotlin/voice/bookOverview/bottomSheet/BottomSheetViewModel.kt
```diff
@@ -23,17 +23,16 @@ class BottomSheetViewModel
   internal fun bookSelected(bookId: Book.Id) {
     this.bookId = bookId
     scope.launch {
-      val items = viewModels.flatMap { it.items(bookId) }
-        .toSet()
-        .sorted()
+      val items = viewModels.sortedBy { it.menuOrder }
```
PaulWoitaschek, line 26: Following YAGNI I'd rely on the order of the enum entries until a more smart ordering is needed.

bookOverview/src/main/kotlin/voice/bookOverview/bottomSheet/BottomSheetViewModel.kt
```diff
@@ -23,17 +23,16 @@ class BottomSheetViewModel
...
       _state.value = EditBookBottomSheetState(items)
     }
   }
 
-  internal fun onItemClick(item: BottomSheetItem) {
+  internal suspend fun onItemClick(item: BottomSheetItem) {
     val bookId = bookId ?: return
     viewModels.forEach {
-      it.onItemClicked(bookId, item)
+      if (it.onItemClicked(bookId, item)) return@forEach
```
PaulWoitaschek, line 35: This short circuit does not improve the performance of a critical path so for simplicity I'd delegate the call to all child view models.
In addition, this would leave the room open for side effects. Multiple view models could participate in the same event.

PR #1343, opened 2022-07-14
Rebuild menu for selected book on orientation change

shirish87:
PR #1341 appears to have an issue with repopulating the menu on orientation change. `EditBookBottomSheetState.items` becomes an empty list. Please review.


PaulWoitaschek:
Thanks for the report and the fix! 

I fixed this on a more general level that will prevent such kind of bugs by retaining the view models themselves in https://github.com/PaulWoitaschek/Voice/commit/d61419f416321f0ab26773c679a4813e8db7c242.

Issue #1447, opened 2022-08-14
Filenames in chapter titles are too long


gullcomb:
### What happened?

In some books, presumably because the relevant metadata is not available, chapter titles display the files' full paths. However, that is often unusable because the display is truncated before showing any chapter-specific information (e.g. `Device:Audiobooks/Very Long Book Title/…`). It would be much more useful if the path was instead relative to the books' directories, or even just the filename without any directory at all. (I don't use subdirectories inside books so those would be the same to me.)

### Version

6.0.7 (via F-droid)

PaulWoitaschek:
Could you attach / send a file by mail that reproduces this issue?

gullcomb:
I can reproduce with just the two attached files (as a "book" with two "chapters"): [Test.zip](https://github.com/PaulWoitaschek/Voice/files/9333929/Test.zip) (courtesy of https://github.com/angband/angband)

If it helps, the audiobooks' folder is on a SD card formatted as removable storage (even though it isn't actually removable) and the phone is running Android 11 (via LineageOS 18.1).

I also did a bit of code diving and the only relevant thing I could find was [this change](https://github.com/PaulWoitaschek/Voice/commit/f00a297c31f94d18887c51130b3a2d38f89122ec#diff-a03355471c266c69da52b962ab5bf55d1c0b7e4c96ca2676cb0f7d400cd6bc19L91-L92). With the files I just uploaded I could see the full chapter title without truncation and the file extension is correctly removed.

In light of the above, my current theory is that `file.lastPathSegment` somehow does not work as intended if the file is on removable storage.

thezeroalpha:
Same issue here, not related directly to removable storage. Even on a device with no SD card, I get the full path, prefixed with `primary:`.

[...]

PR #1464, opened 2022-08-26
Fix chapter names fallback

thezeroalpha:
The chapter name used as a fallback evaluates to a long absolute path,
which is not easy to read and may get truncated. This commit shortens the name. See
https://github.com/PaulWoitaschek/Voice/issues/1447

### DIFF
--- a/scanner/src/main/kotlin/voice/app/scanner/MediaAnalyzer.kt
+++ b/scanner/src/main/kotlin/voice/app/scanner/MediaAnalyzer.kt
@@ -2,6 +2,7 @@ package voice.app.scanner
 
 import android.content.Context
 import android.net.Uri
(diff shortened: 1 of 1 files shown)


thezeroalpha:
I created a DocumentFile from the Uri and obtained the name that way. It seems to be working as it should. 

However, I'm not sure how to write tests for this, as I don't know Kotiln or the architecture of Voice that well. Should the tests be in MediaAnalyzer? And how can I access the metadata of books (specifically chapter names) after the call to `scan(audiobookFolder)`? I could only find chapter IDs in the `bookContentRepo`.

PaulWoitaschek:
Thanks for the contribution 🎉 

As part of https://github.com/PaulWoitaschek/Voice/pull/1465 I simplified the handling by passing the DocumentFile directly.

PR #1488, opened 2022-09-12
Add "audiobook" to cover from internet search

EdwardBetts:
Currently the cover from internet image picker finds images that are taller than they are wide, like the cover of a printed book.

Searching instead for book title + "audiobook cover" returns square covers.


PaulWoitaschek:
Thanks, LGTM!

PR #1647, opened 2022-12-17
Add sleep at end of chapter

tomaThomas:
Implements #1323

I'm not familiar with Kotlin and the android media player, so I'm not sure if this is a good way to implement it (periodically check similar to normal sleep timer). Seems a bit hacky to me.

Also, the UI could be improved, but I don't have a better idea right now.

voice
voice2

Let me know what you think

### DIFF
--- a/playback/src/main/kotlin/voice/playback/PlayerController.kt
+++ b/playback/src/main/kotlin/voice/playback/PlayerController.kt
@@ -141,6 +141,13 @@ class PlayerController
     controller.setPlaybackSpeed(speed)
   }
 
+  fun pauseAtStart() = executeAfterPrepare {
+    it.pause()
+    val bookId = currentBookId.data.first() ?: return@executeAfterPrepare
+    val book = bookRepository.get(bookId) ?: return@executeAfterPrepare
+    it.seekTo(book.currentMark.startMs)
+  }
+
   fun setGain(gain: Decibel) = executeAfterPrepare { controller ->
     controller.sendCustomCommand(CustomCommand.SetGain(gain))
   }
--- a/playback/src/main/kotlin/voice/playback/di/PlaybackModule.kt
+++ b/playback/src/main/kotlin/voice/playback/di/PlaybackModule.kt
@@ -16,6 +16,7 @@ import dagger.Provides
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
@@ -26,6 +27,7 @@ import voice.playback.player.OnlyAudioRenderersFactory
 import voice.playback.player.VoicePlayer
 import voice.playback.player.onAudioSessionIdChanged
 import voice.playback.playstate.PlayStateDelegatingListener
+import voice.playback.playstate.PlayStateManager
 import voice.playback.playstate.PositionUpdater
 import voice.playback.session.LibrarySessionCallback
 import voice.playback.session.PlaybackService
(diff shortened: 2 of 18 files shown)


theAkito:
Great feature, I need that one.

I might look at this, but it seems like the maintainer is currently inactive. 🙁

waot:
There is a new release out, so hopefully it can be incorporated soon :)

PaulWoitaschek:
Thanks for your contribution! The taken approach is a little hacky though. What do you think about attempting the solution that was suggested here? https://github.com/google/ExoPlayer/issues/10262#issuecomment-1128131840

PaulWoitaschek:
The button is fine. I want to rework the UI of that whole thing anyways.

PaulWoitaschek:
Hmm the more I think about it I think the suggested approach can't work with setPauseAtEndOf because exoplayer doesn't know anything about the chapter marks.

playback/src/main/kotlin/voice/playback/PlayerController.kt
```diff
@@ -132,6 +132,11 @@ class PlayerController
     it.seekTo((it.currentPosition - rewind.inWholeMilliseconds.coerceAtLeast(0)))
   }
 
+  fun pauseAtStart() = executeAfterPrepare {
+    it.pause()
+    it.seekTo(0)
```
PaulWoitaschek, line 137: For files containing chapter marks this will seek to the beginning of a file and not to the beginning of a chapter mark

playback/src/main/kotlin/voice/playback/di/PlaybackModule.kt
```diff
@@ -86,18 +88,19 @@ object PlaybackModule {
...
       .build()
       .also { session ->
         scope.launch {
-          sleepTimer.leftSleepTimeFlow
-            .map { it != Duration.ZERO }
-            .distinctUntilChanged()
-            .collect { sleepTimerActive ->
-              sleepTimerCommandUpdater.update(session, sleepTimerActive)
-            }
+          merge(
```
PaulWoitaschek, line 98: I don't think merge is correct here. The latest value is not necessarily the right one. Probably a combine and then conditional logic makes more sense

playback/src/main/kotlin/voice/playback/PlayerController.kt
tomaThomas: I hope that's correct now, seek to the start time of the current chapter mark

PaulWoitaschek:
Just a little heads up: I'm currently very busy and want to first focus on stability before pushing new features. So unfortunatelly this PR has to wait for a while.

tomaThomas:
> Just a little heads up: I'm currently very busy and want to first focus on stability before pushing new features. So unfortunatelly this PR has to wait for a while.

No worries, I'll keep the branch updated because I'm using my own build

obbardc:
Would it make sense to have an option "End of chapter" which sets a timer for the remaining length of the chapter?

e.g. when you press the button it just works out the remaining time left and sets that as the timer length? :-)

tomaThomas:
> Would it make sense to have an option "End of chapter" which sets a timer for the remaining length of the chapter?

I did try that first, but wasn't happy with it. It was not that exact, pausing / skipping backwards / changing playback speed etc. messed it up, and fade out would need to be disabled

obbardc:
Looking great, no hacky timers needed! <3

PaulWoitaschek:
Heads up:
In preparation for merging this PR I've reamped the layout so we can easily find a good UI option for this:
https://github.com/PaulWoitaschek/Voice/pull/2091

PaulWoitaschek:
Wonderful, thank you for the contribution! And sorry for taking ages to merge it 🙈

PaulWoitaschek:
Sorry to tell, but bad news 😢 

The implementation was bugs when files contained multiple chapters. As right now the main branch was not releasable for a while I decided to temporarily revert the change until it's bug free.
https://github.com/PaulWoitaschek/Voice/pull/2264

I'm looking forward for an updated PR 🙏

tomaThomas:
That was the limitation of doing this at the ExoPlayer level - I thought you were aware. I was surprised you merged it in the  first place ^^
I thought about setting a time based sleep timer for the remaining time, but that didn't bother me enough because I don't have such audiobooks. But I will probably deal with it next month when I've got more time

PaulWoitaschek:
Maybe this could be an approach:
https://github.com/google/ExoPlayer/issues/4176#issuecomment-384240744

Issue #1666, opened 2023-02-21
Skip label does not show skip value


sandorex:
### Overview

So changing skip amount is not reflected on the label

### How to reproduce

1. Open settings
2. Change skip amount
3. The label below buttons still show 20s

### Version

6.5.0

### Android Version

31

### Device Model

Xiaomi redmi note 11 pro 5g

sandorex:
This is a duplicate of #1651 , sorry

Issue #1651, opened 2022-12-29
UI inconsistency with Skip amount and Auto rewind


ctntt:
### Overview

If you go to the preferences and change the skip amount (f.e. to 30 seconds) it doesn't change the '20s' to '30s' below the skip buttons in the player. (Maybe one could remove the '20s' below the button and just display the arrows.)

Also if you go to the preferences and click on 'Auto rewind' the text 'Skip amount' is displayed above the slider instead of 'Auto rewind'.

### How to reproduce

1. Open the App
2. Click on the gear icon 
3. In preferences click on Skip amount 
4. Change the Skip amount
5. Close Preferences
6. Click on a book and look at the skip buttons
![Screenshot_20221230](https://user-images.githubusercontent.com/115667885/210020579-a29c863a-6f56-498c-9063-e658226a9283.png)

### Version

6.5.0

### Android Version

12

### Device Model

OnePlus 8t

PaulWoitaschek:
This is fixed in 7.0 (by removing the seek time from the buttons)

PR #1672, opened 2023-02-28
Fix: settings skip amount not reflected in skip buttons

pat721:
Fixes the issues mentioned in  #1666 and #1651 where the skip amount set in the settings is not reflected in the skip buttons of the audiobook player.


PaulWoitaschek:
Thanks for the contribution! This however was already fixed in https://github.com/PaulWoitaschek/Voice/pull/1673

Issue #1656, opened 2023-01-10
Cover cropped in grid layout


FlyingMaddog:
### Overview

The ratio of the cover in grid layout is not 1:1

### How to reproduce

Open app
switch to grid view

### Version

6.4.1

### Android Version

13

### Device Model

SM-S901B

PaulWoitaschek:
That's actually on purpose so more books can fit on the screen.

PR #1678, opened 2023-03-03
Fix aspect ratio of images in grid layout

pat721:
Fixes the aspect ratio to 1:1 when in grid layout as requested by #1656.


PaulWoitaschek:
Thanks for the PR! That's actually on purpose so more books can fit on the screen.

PR #2211, opened 2023-12-04
Sort new audiobooks by new

newhinton:
This allows the "new" category to order the books by their date added, which is the expected ordering for new content.

### DIFF
--- a/bookOverview/src/main/kotlin/voice/bookOverview/overview/BookOverviewCategory.kt
+++ b/bookOverview/src/main/kotlin/voice/bookOverview/overview/BookOverviewCategory.kt
@@ -16,7 +16,7 @@ enum class BookOverviewCategory(
   ),
   NOT_STARTED(
     nameRes = StringsR.string.book_header_not_started,
-    comparator = BookComparator.ByName,
+    comparator = BookComparator.ByLastAdded,
   ),
   FINISHED(
     nameRes = StringsR.string.book_header_completed,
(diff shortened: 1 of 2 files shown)


PaulWoitaschek:
I'm unsure what a good solution is here. Initially Voice was implemented like you wrote, but for new users this is getting messy. Because on their initial import, everything looks completely random, though users rather expect to search by title here.

newhinton:
Hmm yeah, i see the initial issue. However, i think with ongoing usage it makes way more sense to sort by new. I would assume that adding (and therefore finding) the newly added content to be the more important task.

newhinton:
Maybe it is possible to direct the initial scan to add the content by alphabet, so that we get an initial alphabetical order? This would solve both issues.

PaulWoitaschek:
That would complicate the code by quite a bit.

If we take a step back, what's the actual use case you're trying to solve?
It's probably this right?

You add a new book to your collection. Now you want to play it. 

We could solve this from the product side:

## Rename the "New" category to "Unplayed".

That would actually make sense because then the three categories would form some kind of a trinity:
* Current
* Not Played
* Finished Playing

## Mark "New" books
Then in the "Not Played" section, we could move books which were added within the latest 24 hours:
1. To the top
2. Within these, sort them by name
3. Add a small chip labeling them as "new"

WDYT?

newhinton:
That is actually a great idea!

Though one thing:
When "New" books become "normal" ones, they will likely be organized alphabetically again. That *might* be confusing again, but should not be too bad. 

I dont know if this would introduce more code again, but is it possible to keep the "new" label, until more books are added? I guess that would only be a nice to have, but this way, the latest added books are always "new", until new content is added. Though i dont know if it is worth the effort.

Issue #2025, opened 2023-07-02
Incorrect Album/Book/Folder name


40ZqTYY4ag3tAO4hdlxR1:
### Overview

Since version 8.0.0 the "Name" assigned when a new folder is added is taken from the first file in the folder (oor the ID3 tag of the first file? ), or the first chapter in the case of a m4a. If the first file does not have any metadata, the filename is used.

Expected/Previous behavior: The "Name" assigned was the folder name.

In the images below, ,even though the folder is named "Wasteland" , the name assigned is ch1, which is the name of the first chapter of the m4a file. The name "Wasteland" in the screen shot was added manually.

![](https://i.ibb.co/dMGkJRq/A.gif)  ![](https://i.ibb.co/WKwn4jQ/B.gif)  ![](https://i.ibb.co/FWN51Js/C.gif)

### How to reproduce

Version 8.1.2

As posting the actual Wasteland book here is a no, no... you can test it with the attached file [Test_Book.zip](https://github.com/PaulWoitaschek/Voice/files/11929409/Test_Book.zip) . The folder is called Test_Book , the file contained in it is called **beep_file.m4a** . **beep_file.m4a** has 10 chapters each 30 seconds long.

The expected behavior : the album name should be the folder name or file name in case of a file with chapters or the album-title from the metadata.

Current behavior : 
![](https://i.ibb.co/4d2qm9f/Test-Book-A.gif) ![](https://i.ibb.co/qnYrZnS/Test-Book-B.gif) 

### Version

8.1.2

### Android Version

10

### Device Model

SM

PaulWoitaschek:
Why is this completed?

40ZqTYY4ag3tAO4hdlxR1:
It's such a minor thing. Just add the album name manually. This is not a podcast feed that gets updated every few hours.

The problem has no negative effects on the user experience or the app's operation.

Your work on Voice, which has been great for years, is greatly appreciated.

PaulWoitaschek:
Thanks. Bugs need to be fixed :)

RecentRunner:
Agreed. Has there been any towards fixing this?

Vuizur:
@RecentRunner I think I found the reason and tried to fix it in this PR: https://github.com/PaulWoitaschek/Voice/pull/2228

PR #2228, opened 2023-12-17
Fix bug where file name instead of folder name is chosen as audiobook title

Vuizur:
This should fix https://github.com/PaulWoitaschek/Voice/issues/2025 (at least I hope).

I tested it with two books: One with ID3 tags and one without, the tagged book was displayed as before, the book without tags now has the book folder name as expected (instead of the name of the first file.) 

Thanks a lot for developing this great application!

### DIFF
--- a/scanner/src/main/kotlin/voice/app/scanner/BookParser.kt
+++ b/scanner/src/main/kotlin/voice/app/scanner/BookParser.kt
@@ -59,7 +59,7 @@ class BookParser
         author = analyzed?.author,
         lastPlayedAt = migrationSettings?.lastPlayedAtMillis?.let(Instant::ofEpochMilli)
           ?: Instant.EPOCH,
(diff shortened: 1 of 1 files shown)


PaulWoitaschek:
@Vuizur 

Thanks for your contribution! There are right now three PR's which touch this.

Could you check if this one fixes it for you?
#2261

Vuizur:
Thanks a lot for looking into it 😀. Unfortunately my GPU is broken currently, so my Android emulator bluescreens, and I didn't want to uninstall the app on my smartphone (to not delete my progress). But I will report with the next release if it works 👍

PaulWoitaschek:
Can you check the latest release?
https://github.com/PaulWoitaschek/Voice/releases/download/8.2.0/app-release.apk

Vuizur:
Yes, it seems to be fixed now. Thanks a lot!

Issue #2171, opened 2023-10-22
Podcast titles are all the same


sigmike:
### Overview

Hello,
I use Voice to listen to podcasts I download. When I add a folder with all the podcast tracks it works fine but all the tracks have the same title.
Comparing these titles with `mediainfo` of the files it looks like Voice uses the "Performer" attribute as the first title, and the "Album" as the second title. It would be more relevant to use the "Track name" as second title.

### How to reproduce

- Download [this file from the 80.000 hours podcast](https://media.transistor.fm/cbcc6bbe/278552ba.mp3) into a folder
- Add the folder
- The title is the album name `80,000 Hours Podcast` whereas it should be the track name `#167 – Seren Kell on the research gaps holding back alternative proteins from mass adoption`

### Version

8.1.3

### Android Version

8

### Device Model

Galaxy A3

PaulWoitaschek:
Closing as completed, please report if it still happens

wrenix:
I got the same problem with 8.2.0

PR #2253, opened 2024-01-15
fix: use title tag as book name

FineFindus:
Fixes an issue where the album tag was displayed instead of the title. This can be an issue with audio files that are part of a larger series, where the album is set to the series title instead of the episode title.

| Before                                                                                                                                             	| After                                                                                                                                                        	|
|----------------------------------------------------------------------------------------------------------------------------------------------------	|--------------------------------------------------------------------------------------------------------------------------------------------------------------	|
| Listitem showing 'Auphonic Examples' as the title 	| Listitem showing 'Auphonic Chapter Marks Demo' as the title 	|

For example, [Amberol](https://apps.gnome.org/Amberol/) correctly uses the title tag.
Audio file in Amberol with correct title

Closes https://github.com/PaulWoitaschek/Voice/issues/2171

I was wondering what `chapterName` actually does. I couldn't find any use for it in the UI.

### DIFF
--- a/app/src/androidTest/kotlin/voice/app/misc/MediaAnalyzerInstrumentationTest.kt
+++ b/app/src/androidTest/kotlin/voice/app/misc/MediaAnalyzerInstrumentationTest.kt
@@ -36,7 +36,7 @@ class MediaAnalyzerInstrumentationTest {
     analyze(R.raw.intact) shouldBe MediaAnalyzer.Metadata(
       duration = 119040,
       author = "Auphonic",
-      bookName = "Auphonic Examples",
+      bookName = "Auphonic Chapter Marks Demo",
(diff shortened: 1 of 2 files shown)


PaulWoitaschek:
Could you check this one?

https://github.com/PaulWoitaschek/Voice/pull/2261

Should bring the best of both worlds.

FineFindus:
Hmm, I completely overlooked the use case of having chapters as individual audio files, since I always use embedded chapters. I tested https://github.com/PaulWoitaschek/Voice/pull/2261 and it works, great idea. So I'm closing this one in favor of it :)

PaulWoitaschek:
Yep, it's quite complicated to do stuff with the various types of use cases in mind and not breaking use cases 🤷

PR #2261, opened 2024-01-19
For single-file audiobooks prefer the title over the album for the book name

PaulWoitaschek:
Closes #2253 #2171

### DIFF
--- a/scanner/src/main/kotlin/voice/app/scanner/BookParser.kt
+++ b/scanner/src/main/kotlin/voice/app/scanner/BookParser.kt
@@ -51,15 +51,31 @@ class BookParser
 
       val migratedPlaybackPosition = migrationSettings?.let { findMigratedPlaybackPosition(it, chapters) }
 
+      buildList {
+        add(null)
+        add("")
+      }.filterNotNull().firstOrNull()
       BookContent(
         id = id,
         isActive = true,
         addedAt = migrationMetaData?.addedAtMillis?.let(Instant::ofEpochMilli)
           ?: Instant.now(),
-        author = analyzed?.author,
+        author = analyzed?.artist,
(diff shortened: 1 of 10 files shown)


scanner/src/main/kotlin/voice/app/scanner/BookParser.kt
```diff
@@ -51,15 +51,31 @@ class BookParser
...
         addedAt = migrationMetaData?.addedAtMillis?.let(Instant::ofEpochMilli)
           ?: Instant.now(),
-        author = analyzed?.author,
+        author = analyzed?.artist,
         lastPlayedAt = migrationSettings?.lastPlayedAtMillis?.let(Instant::ofEpochMilli)
           ?: Instant.EPOCH,
-        name = migrationMetaData?.name ?: analyzed?.bookName ?: analyzed?.chapterName ?: file.bookName(),
+        name = buildList {
+          add(migrationMetaData?.name)
+          if (chapters.size == 1) {
```
balki, line 68: Does `chapters.size == 1` mean single file book? I see in some podcast episode mp3s, there are multiple chapters embedded in a single mp3. Those should also have title as the book name or may be '{album} - {tittle}'.

balki: @PaulWoitaschek 
```suggestion
          if (!file.isDirectory) {
```

PaulWoitaschek: One chapter = one file already. Every chapter itself can contain chapter marks. 
It's not the best naming currently

balki: I see. Thanks for clarifying! Looking forward for the next release.

PR #2279, opened 2024-01-27
Enable receiving MEDIA_BUTTON intent

ASGusev:
In #2198 I have reported that the app fails to continue playback from the notifiation after being unloaded and suggested adding an intent receiver to the manifest. In this pull request I suggest merging this receiver


PaulWoitaschek:
Thanks for your PR! However that's already done here: https://github.com/PaulWoitaschek/Voice/blob/main/playback%2Fsrc%2Fmain%2FAndroidManifest.xml

PR #2280, opened 2024-01-27
Fix playback continuation on external controller command involving search (through the onPlayFromSearch)

ASGusev:
In #2215 I have mentioned a limitation of external controller support and described a possible solution. So, here I propose some changes to widen the support for external controllers.


PaulWoitaschek:
Hmm, I'm not sure if this is the correct solution. The code is only a weird workaround. Usually connecting to a media controller should not prepare anything. Instead, the client connecting to a session should issue preparing of the corresponding media.

ASGusev:
I am considering the scenario when the controller connects and issues a play command. Most apps continue to play the media they played last, Voice now picks a book from those which have not been started yet. Do you mean that there should be an extra action between the connection and play command initiating loading the media? Or should it be loaded on service start?

PaulWoitaschek:
I'm hesitant on this change; especially around re-injecting the service. I think this is not how playback resumption is mean to work with media3. 

I've raised a question here: https://github.com/androidx/media/issues/1064

Could you add the details from your original discussion to this issue so the information is less spread and media3 developers have an easier time?

ASGusev:
I have moved the position loading from `onConnect` to the PlaybackService's `onCreate` so that the service always has the position  ready

PaulWoitaschek:
I would assume that this would lead to a permanent notification which has been the case in the past. 
Can you check that? (Ie there were a lot of users complaining that Voice is always there, even though the user did not launch it. )

ASGusev:
Well, now I see some notifications but not permanent: they sometimes appear on app launch (is is a problem?) and swiping it away from the recent app list (trying to figure out why it happens and how to filter it out).
Do you think it can be fine to check the player state before setting `playWhenReady = true` in `VoicePlayer.play` and load state if the state is idle? Or to load the position on service start and to prepare the player if it is asked to play from idle state?

PaulWoitaschek:
Can you check Marcs answer? From what he explained it looks like this change should not be necessary and another controller should just connect and then call play()

ASGusev:
Yes, I have explored what's happening again and see another solution. SimpleWear, with which I face the issue, uses legacy controller's `onPlayFromSearch` method without query for which, according to android reference the player app is expected to make "a smart choice about what to play". By the media3 session this call gets delegated to `getSearchResults`. In Voice it is processed by `BookSearchHandler.searchUnstructured`. This method already returns the last played book when called with null as query, but during `onPlayFromSearch` processing it receives an empty string and returns an arbitrary book. So, now I suggest adjusting the check in `searchUnstructured` so that it also returns the last book when called with an empty string.

playback/src/main/kotlin/voice/playback/session/search/BookSearchHandler.kt
```diff
@@ -44,7 +44,7 @@ class BookSearchHandler
 
   // Look for anything that might match the query
   private suspend fun searchUnstructured(query: String?): Book? {
-    if (query != null) {
+    if (query != null && query != "") {
```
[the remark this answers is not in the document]
ASGusev: Switched

PaulWoitaschek:
Very cool!

Could you merge main and update the PR title to describe what it fixed?

ASGusev:
Will it do?

playback/src/main/kotlin/voice/playback/session/PlaybackService.kt
```diff
@@ -57,14 +57,13 @@ class PlaybackService : MediaLibraryService() {
...
-    return session.takeUnless { session ->
-      session.invokeIsReleased
-    }.also {
-      if (it == null) {
-        Logger.e("onGetSession returns null because the session is already released")
-      }
-    }
+  override fun onGetSession(controllerInfo: MediaSession.ControllerInfo): MediaLibrarySession {
+    if (session.invokeIsReleased)
+      rootComponentAs<PlaybackComponent.Provider>()
```
PaulWoitaschek, line 62: And are you sure this is necessary? I'm hesitant to merge this because this will literally create new instances of the whole player stuff - while not releasing the old things.
I fear that this might lead to several players playing at the same time under some circumstances

ASGusev: A new session is only created in a situation when the service is not destroyed after a `release` call. I face this situation when I run Voice from SimpleWear and then close it. Now in such a situation playback cannot be started even by the play button in the player activity (or the play button in the books list). I think this behaviour is not more desirable.
By the way, in the comment in `onTaskRemoved` you have written that releasing the session is required to prevent showing a dead notification. Later, in a comment to my previous PR, you've mentioned having fixed the notification being unresponsive. So, is the dead notification still an issue, and is the `release` call in `onTaskRemoved` still required? It seems to me that only calling `release` from `onDestroy` would obviate the sole need to check if the session state.

ASGusev: And which conditions do you mean? Is that `onGetSession` being called from another thread than `onCreate` that does not see the session created there?

ASGusev: Also, I don't quite understand what you mean by 'not releasing the old stuff' - a new session is only created if the old one has already been released.

ASGusev: Any updates?:)

PaulWoitaschek:
Thanks for your MR!

I did cherry-pick the commit that changes searchUnstructured.
I do not want to merge the change that nulls out the underlying session as that's a workaround without a test case which can lead to unpredictable behaviour

ASGusev:
Great! Maybe I can provide a case of the issue with null returned from onGetSession?

PaulWoitaschek:
I think that should be implemented in the upstream media repository as it is not specific to voice but should affect all apps.

PR #2289, opened 2024-02-04
Add mpga extension as supported audioformat

balki:
Some audio files have this extension. When I rename to `mp3`, they work fine with Voice.


PaulWoitaschek:
Thanks! <3

Issue #2103, opened 2023-08-30
Feature request - Backup and restore option


W1zardK1ng:
### Overview

Please add an option to export current app setting to a json file or something else and restore the book listen position and cover photo.

### How to reproduce

this will help to restore the app after a phone reset or app reinstall.

### Version

8.1.3

### Android Version

Android 13

### Device Model

Pixel 6a

polyzen:
Duplicate of https://github.com/PaulWoitaschek/Voice/discussions/1668

PR #2302, opened 2024-02-15
Implement Backup &  Restore

jaredly:
as requested in https://github.com/PaulWoitaschek/Voice/discussions/1668 and https://github.com/PaulWoitaschek/Voice/issues/2103

I'm quite interested in being able to transfer read information, as I'm switching to a new phone and am in the middle of quite a few books 😅

NB I haven't written very much kotlin, so it's entirely likely that I've organized things weirdly. Let me know how you'd like things to be rearranged! I was pretty much just blundering around to get it to compile & work 🙃

But at the end of the day, it does work! I based my implementation fairly loosely on https://stackoverflow.com/a/73205884/290784. I'm exporting the whole sqlite file (including journaling files) as a .zip archive, and then to restore, I overwrite whatever sqlite database exists & restarting the app.
(It might be worth putting up an alert or something saying "Restoring will delete all your current data!" but I wasn't sure how to make an alert)

Here's a video of it in action:
https://github.com/PaulWoitaschek/Voice/assets/112170/d512e8b8-2762-4945-8a54-11ac84c3c952


gradle.properties
```diff
@@ -14,3 +14,4 @@ org.gradle.unsafe.configuration-cache=true
 voice.signing.play=false
 kotlin.daemon.useFallbackStrategy=false
 org.gradle.kotlin.dsl.allWarningsAsErrors=true
+org.gradle.jvmargs=-Xmx2g
```
jaredly, line 17: this was required for me to be able to build the project the first time, but I can back it out if you want

jgilmore:
Yeah, good to know this isn't implemented yet. Too bad I can't see the reading positions on my old phone, since the SD card that had the books on it has been moved to the new phone already.

So I'm SOL I guess.

So I guess I should ask if this pull request backs up and restores the reading positions even for books that are no longer available on the device?

jaredly:
@jgilmore I believe it does

eturk1:
Thanks for the effort to implement a manual backup/restore! 

First, is Android's Auto Backup implemented? 
https://developer.android.com/identity/data/autobackup
Moving to a new phone should work automatically

jaredly:
doesn't seem to have worked when I moved to a new phone the most recent time

eturk1:
> doesn't seem to have worked when I moved to a new phone the most recent time

Is it set in the manifest? 
How do we debug auto backup, check for the backup for in Drive?

PaulWoitaschek:
Thanks for the contribution!

I think generally this needs a different approach. 
Backing up the pure SQL database is very error prone, especially around database scheme updates. 
Also there is a fundamental problem: 
Voice uses uris for the files. They are different from phone to phone. 
So even if this might work for some cases it can not work for all phones.

valerie-garoux:
> I think generally this needs a different approach. Backing up the pure SQL database is very error prone, especially around database scheme updates. Also there is a fundamental problem: **Voice uses uris for the files.** They are different from phone to phone. So even if this might work for some cases it can not work for all phones.

Can't we imagine a dialog when importing the backup to select the folders/files back to rewrite the URIs if they're not ✨ automagically ✨ retrieved?
I'm moving from a phone to another at the moment and the perspective of redoing all of the sorting on the Voice app is a bit hard...

PR #2329, opened 2024-03-09
For the book name, prefer the album over the title

PaulWoitaschek:


balki:
Why is this changed again for single file books (also for podcast files)? Can you please provide a config option to keep the previous behavior if it breaks some other use case?

Issue #2216, opened 2023-12-10
Extremely High Battery Usage 


hchasens:
### Overview

According to my battery manager, Voice uses almost 50% of my phone's battery. I know this is in relation to how much battery I use, but it still seems excessive. Is there any way we can reduce background processing without stopping playback? I'm using Bluetooth LE so it's not the speakers. 

### How to reproduce

Open App
Click Play
Lock Phone
Wait a While
Check Battery Usage

### Version

8.1.3

### Android Version

Graphene OS build 2023129

### Device Model

Pixel 8 Pro

PR #2413, opened 2024-05-16
feat(player): enable audio offloading

FineFindus:
Enables the [AudioOffloading](https://developer.android.com/media/media3/exoplayer/battery-consumption#audio-playback) feature of the Exoplayer, which reduces power consumption and hopefully helps with issues like https://github.com/PaulWoitaschek/Voice/issues/2216.

I have not had time to fully test this yet.


PaulWoitaschek:
Iirc it does not work with skip silence and speed adjustments

FineFindus:
> Iirc it does not work with skip silence and speed adjustments

Hmm, that would be a bummer? I thought `setIsGaplessSupportRequired` and `setIsSpeedChangeSupportRequired` would be enough to be compatible with both features, it is also used in this configuration in other audio players that support both skip silence and speed change.

PaulWoitaschek:
Ooh you are right! I did not check the pr properly, my bad. 

I think these changes are dangerous nevertheless. Wdyt about a setting: Experimental audio Player
In the settings menu which enables these newer yet interested things?

FineFindus:
I'm generally not a fan of settings for various reasons (IMHO they're often bad for both the user and the maintainer). Maybe it could be added as an experimental setting first, and after some time, if there are no problems, move to it being enabled by default?

PaulWoitaschek:
Yep that was my idea

FineFindus:
I've been trying to implement the settings, but due to me being inexperienced/not understanding the architecture/not being an android dev, I've had some trouble getting it to work. Would appreciate some help/pointers, otherwise I will take another look when I have some more time.

PR #2668, opened 2025-01-05
Add sleep at end of chapter

tomaThomas:
Hi, I'm back with an implementation of the sleep at end of chapter that works with chapter marks in a single media file.
Stops between media files are handled in the `onMediaItemTransition` listener . Additionally, for every chapter mark a callback is registered, that stops the playback if sleep at end of chapter is active when the mark is reached.

### DIFF
--- a/playback/src/main/kotlin/voice/playback/PlayerController.kt
+++ b/playback/src/main/kotlin/voice/playback/PlayerController.kt
@@ -148,9 +148,7 @@ class PlayerController
 
   fun pauseAtStart() = executeAfterPrepare {
     it.pause()
-    val bookId = currentBookStoreId.data.first() ?: return@executeAfterPrepare
-    val book = bookRepository.get(bookId) ?: return@executeAfterPrepare
-    it.seekTo(book.currentMark.startMs)
+    it.seekTo(0)
   }
 
   fun pauseAtTime(ms: Long) = executeAfterPrepare {
--- a/playback/src/main/kotlin/voice/playback/PlayerController.kt
+++ b/playback/src/main/kotlin/voice/playback/PlayerController.kt
@@ -29,6 +29,7 @@ import voice.playback.session.toMediaIdOrNull
 import javax.inject.Inject
 import kotlin.coroutines.coroutineContext
 import kotlin.time.Duration
+import kotlinx.coroutines.delay
 
 class PlayerController
 @Inject constructor(
@@ -148,6 +149,7 @@ class PlayerController
 
   fun pauseAtStart() = executeAfterPrepare {
     it.pause()
+    delay(100)
     it.seekTo(0)
   }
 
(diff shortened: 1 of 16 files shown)


pktiuk:
@PaulWoitaschek Do you plan to review this change?

playback/src/main/kotlin/voice/playback/PlayerController.kt
```diff
@@ -146,6 +146,18 @@ class PlayerController
     controller.setPlaybackSpeed(speed)
   }
 
+  fun pauseAtStart() = executeAfterPrepare {
```
PaulWoitaschek, line 149: It's working great with chapters! However the following scenario doesn't work as expected:

- Have two mp3's with chapters
- Seek to the last chapter of the first mp3
- Enable EOC
- Play

The expected behavior would now be to stop at the beginning of the second mp3. However it rewinds all the way back to the beginning of the first mp3

playback/src/main/kotlin/voice/playback/playstate/PlayStateManager.kt
```diff
@@ -25,4 +25,16 @@ constructor() {
     Playing,
     Paused,
   }
+
+  // Sleep at eoc state
+  private val _sleepAtEoc = MutableStateFlow(false)
```
PaulWoitaschek, line 30: To me this conceptually fits better in the SleepTimer itself. What do you think about moving that there?

tomaThomas: Agreed

playback/src/main/kotlin/voice/playback/PlayerController.kt
tomaThomas: Getting the chapter mark wasn't necessary anymore because the method is only called in `onMediaItemTransition`
ExoPlayer seems to call `onMediaItemTransition` before it updates its internal state - for now I just added a delay. Cannot think of a better solution other than polling right now

tomaThomas: Thinking about it, I never had a problem for files without chapter marks, so maybe a chapter mark at the beginning of a file are causing this problem

PaulWoitaschek:
I don't want to merge this PR without a test-suite covering it.

In #2836 I'm currently setting up the infrastructure to verify that the sleep timer is behaving as expected. Once that's merged, we can add this PR's logic changes to that test case.

tomaThomas:
That's great!
Also, I want to look for a better solution than that delay

PaulWoitaschek:
@tomaThomas 
Could you create a branch on your repo and work there instead of main? I wanted to push some minor things but can't because even though It's set as "allow edits from maintainers" - this doesn't work for the main branch.

PaulWoitaschek:
> I think I can't change the base branch of this PR, so I created a new one, #2838 

If I click on the branch I can select another one, did you try that?

tomaThomas:
> If I click on the branch I can select another one, did you try that?

Didn't find the option on the web version, but still, the app only allows me to change the target branch in your repo

PR #2673, opened 2025-01-14
Automatically start sleep timer on time schedule

happy-turtle:
Referring to [this feature request](https://github.com/PaulWoitaschek/Voice/discussions/1743). I implemented a new setting for the sleep timer where one can set a start and an end time. If the current local time is in-between the two, the sleep timer will start automatically on playback. I am using this feature on my own device already, but of course I'm open for any suggestions on UI changes or similar.
sleep timer screen with sleep timer schedule turned off sleep timer screen with sleep timer schedule turned on open time picker dialog

### DIFF
--- a/app/src/main/kotlin/voice/app/injection/PrefsModule.kt
+++ b/app/src/main/kotlin/voice/app/injection/PrefsModule.kt
@@ -15,9 +15,11 @@ import voice.bookOverview.BookMigrationExplanationQualifier
 import voice.bookOverview.BookMigrationExplanationShown
 import voice.common.AppScope
 import voice.common.BookId
+import voice.common.autoSleepTimer.AutoSleepTimer
 import voice.common.grid.GridMode
 import voice.common.pref.AuthorAudiobookFoldersStore
 import voice.common.pref.AutoRewindAmountStore
+import voice.common.pref.AutoSleepTimerStore
 import voice.common.pref.CurrentBookStore
 import voice.common.pref.DarkThemeStore
 import voice.common.pref.FadeOutStore
@@ -30,6 +32,7 @@ import voice.common.pref.SingleFolderAudiobookFoldersStore
 import voice.common.pref.SleepTimeStore
 import voice.common.serialization.UriSerializer
 import voice.datastore.VoiceDataStoreFactory
+import java.time.LocalTime
 import javax.inject.Singleton
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.minutes
@@ -108,6 +111,21 @@ object PrefsModule {
       serializer = Duration.serializer(),
       fileName = "sleepTime2",
       defaultValue = 20.minutes,
+    )
+  }
+
+  @Provides
+  @Singleton
+  @AutoSleepTimerStore
+  fun provideAutoSleepTimerPreference(factory: VoiceDataStoreFactory): DataStore<AutoSleepTimer> {
+    return factory.create(
+      AutoSleepTimer.serializer(),
+      AutoSleepTimer(
+        enabled = false,
+        startTime = LocalTime.of(22, 0),
+        endTime = LocalTime.of(6, 0),
+      ),
+      "autoSleepTimer",
     )
   }
 
--- a/settings/src/main/kotlin/voice/settings/SettingsViewModel.kt
+++ b/settings/src/main/kotlin/voice/settings/SettingsViewModel.kt
@@ -13,14 +13,17 @@ import voice.common.AppInfoProvider
 import voice.common.DARK_THEME_SETTABLE
 import voice.common.DispatcherProvider
 import voice.common.MainScope
+import voice.common.autoSleepTimer.AutoSleepTimer
 import voice.common.grid.GridCount
 import voice.common.grid.GridMode
 import voice.common.navigation.Destination
 import voice.common.navigation.Navigator
 import voice.common.pref.AutoRewindAmountStore
+import voice.common.pref.AutoSleepTimerStore
 import voice.common.pref.DarkThemeStore
 import voice.common.pref.GridModeStore
 import voice.common.pref.SeekTimeStore
+import java.time.LocalTime
 import javax.inject.Inject
 
 class SettingsViewModel
@@ -35,6 +38,8 @@ class SettingsViewModel
   private val appInfoProvider: AppInfoProvider,
   @GridModeStore
   private val gridModeStore: DataStore<GridMode>,
+  @AutoSleepTimerStore
+  private val autoSleepTimer: DataStore<AutoSleepTimer>,
   private val gridCount: GridCount,
   dispatcherProvider: DispatcherProvider,
 ) : SettingsListener {
@@ -48,6 +53,13 @@ class SettingsViewModel
     val autoRewindAmount by remember { autoRewindAmountStore.data }.collectAsState(initial = 0)
     val seekTime by remember { seekTimeStore.data }.collectAsState(initial = 0)
     val gridMode by remember { gridModeStore.data }.collectAsState(initial = GridMode.GRID)
+    val autoSleepTimer by remember { autoSleepTimer.data }.collectAsState(
+      initial = AutoSleepTimer(
+        enabled = false,
+        startTime = LocalTime.of(22, 0),
+        endTime = LocalTime.of(6, 0),
+      ),
+    )
     return SettingsViewState(
       useDarkTheme = useDarkTheme,
       showDarkThemePref = DARK_THEME_SETTABLE,
@@ -60,6 +72,7 @@ class SettingsViewModel
         GridMode.GRID -> true
         GridMode.FOLLOW_DEVICE -> gridCount.useGridAsDefault()
       },
+      autoSleepTimer = autoSleepTimer,
     )
   }
 
@@ -135,5 +148,37 @@ class SettingsViewModel
   override fun openTranslations() {
     dismissDialog()
     navigator.goTo(Destination.Website("https://hosted.weblate.org/engage/voice/"))
+  }
+
+  override fun setAutoSleepTimer(checked: Boolean) {
+    mainScope.launch {
+      autoSleepTimer.updateData { currentPrefs ->
+        currentPrefs.copy(enabled = checked)
+      }
+    }
+  }
+
+  override fun setAutoSleepTimerStart(
+    hour: Int,
+    minute: Int,
+  ) {
+    val time = LocalTime.of(hour, minute)
+    mainScope.launch {
+      autoSleepTimer.updateData { currentPrefs ->
+        currentPrefs.copy(startTime = time)
+      }
+    }
+  }
+
+  override fun setAutoSleepTimerEnd(
+    hour: Int,
+    minute: Int,
+  ) {
+    val time = LocalTime.of(hour, minute)
+    mainScope.launch {
+      autoSleepTimer.updateData { currentPrefs ->
+        currentPrefs.copy(endTime = time)
+      }
+    }
   }
 }
--- a/settings/src/main/kotlin/voice/settings/views/AutoSleepTimerSetting.kt
+++ b/settings/src/main/kotlin/voice/settings/views/AutoSleepTimerSetting.kt
@@ -1,0 +1,51 @@
+package voice.settings.views
+
+import androidx.compose.foundation.layout.Row
+import androidx.compose.material3.LocalContentColor
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.material3.TimePickerState
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
+import java.time.LocalTime
+import java.time.format.DateTimeFormatter
+import java.time.format.FormatStyle
+
+@Composable
+internal fun AutoSleepTimerSetting(
+  autoSleepTimer: Boolean,
+  time: LocalTime,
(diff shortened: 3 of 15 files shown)


PaulWoitaschek:
Nice!

I think the logic would make more sense if it was a little different:

- Add that as a setting to the settings menu 
- Two rows entries, one for the checkbox and dscription, and one for setting the timing
- When the checkbox is is checked, the timing entry gets greyed out

Then it's clearer and more consistent that this is a global setting, affecting all books.

playbackScreen/src/main/kotlin/voice/playbackScreen/BookPlayViewModel.kt
```diff
@@ -182,9 +189,27 @@ class BookPlayViewModel
...
+      val startTime = LocalTime.parse(autoSleepTimerStart.value)
+      val endTime = LocalTime.parse(autoSleepTimerEnd.value)
+      if (isCurrentTimeInRange(startTime, endTime)) {
+        sleepTimer.setActive()
+      }
+    }
     player.playPause()
   }
 
+  private fun isCurrentTimeInRange(startTime: LocalTime, endTime: LocalTime): Boolean {
```
PaulWoitaschek, line 202: I'd move this logic into a dedicate class and unit-test it thoroughly

settings/src/main/kotlin/voice/settings/views/AutoSleepTimerSetting.kt
```diff
@@ -0,0 +1,48 @@
...
+  autoSleepTimer: Boolean,
+  time: String,
+  label: String,
+  setAutoSleepTime: (Int, Int) -> Unit,
+) {
+  val shouldShowTimePicker = remember { mutableStateOf(false) }
+  CompositionLocalProvider(
+    LocalContentColor provides
+      if (autoSleepTimer) MaterialTheme.colorScheme.onSurface
+      else MaterialTheme.colorScheme.onSurface.copy(alpha = 0.38f),
```
PaulWoitaschek, line 24: You can follow the `AppVersion` composable to achieve a consistent disabled look

settings/src/main/kotlin/voice/settings/views/Settings.kt
```diff
@@ -136,6 +148,33 @@ private fun Settings(
...
+        )
+        ListItem(
+          leadingContent = { Spacer(Modifier.size(24.dp)) },
+          headlineContent = {
+            Row(
+              horizontalArrangement = Arrangement.spacedBy(10.dp),
+              verticalAlignment = Alignment.CenterVertically,
+            ) {
+              AutoSleepTimerSetting(
+                viewState.autoSleepTimer,
```
PaulWoitaschek, line 163: This should be a localized time format and not the iso format

settings/src/main/kotlin/voice/settings/views/TimePickerDialog.kt
```diff
@@ -0,0 +1,47 @@
...
+
+@Composable
+fun TimePickerDialog(
+  initialHour: Int,
+  initialMinute: Int,
+  onConfirm: (TimePickerState) -> Unit,
+  onDismiss: () -> Unit,
+) {
+  val timePickerState = TimePickerState(
+    initialHour,
```
PaulWoitaschek, line 20: named args pls

app/src/main/kotlin/voice/app/injection/PrefsModule.kt
```diff
@@ -78,6 +80,27 @@ object PrefsModule {
...
+  @Singleton
+  @Named(PrefKeys.AUTO_SLEEP_TIMER)
+  fun provideAutoSleepTimerPreference(prefs: AndroidPreferences): Pref<Boolean> {
+    return prefs.boolean(PrefKeys.AUTO_SLEEP_TIMER, false)
+  }
+
+  @Provides
+  @Singleton
+  @Named(PrefKeys.AUTO_SLEEP_TIMER_START)
+  fun provideAutoSleepTimerStartPreference(prefs: AndroidPreferences): Pref<String> {
```
PaulWoitaschek, line 93: I would use a single @Serializable data class for the sleep pref instead of three losely typed ones

settings/src/main/kotlin/voice/settings/SettingsViewModel.kt
```diff
@@ -119,4 +132,18 @@ class SettingsViewModel
     dismissDialog()
     navigator.goTo(Destination.Website("https://hosted.weblate.org/engage/voice/"))
   }
+
+  override fun toggleAutoSleepTimer(checked: Boolean) {
```
PaulWoitaschek, line 136: I would not name it toggle. Toggle implies that it's being flipped but through the checked boolean it's already specifying the state

settings/src/main/kotlin/voice/settings/views/Settings.kt
```diff
@@ -49,6 +55,9 @@ private fun SettingsPreview() {
     dialog = null,
     appVersion = "1.2.3",
     useGrid = true,
+    autoSleepTimer = false,
+    autoSleepTimeEnd = "",
```
PaulWoitaschek, line 59: Can we use something more realistic for the preview?

settings/src/main/kotlin/voice/settings/views/Settings.kt
```diff
@@ -136,6 +148,33 @@ private fun Settings(
...
         }
+        AutoSleepTimerRow(
+          viewState.autoSleepTimer,
+          listener::toggleAutoSleepTimer,
+        )
+        ListItem(
+          leadingContent = { Spacer(Modifier.size(24.dp)) },
+          headlineContent = {
+            Row(
+              horizontalArrangement = Arrangement.spacedBy(10.dp),
```
PaulWoitaschek, line 159: I'd follow the standard 4dp grid here

PaulWoitaschek (changes requested):
Nice, great to see this started! <3

happy-turtle:
Thank you for taking the time to thoroughly review this! I will look into your feedback.

PaulWoitaschek:
@happy-turtle 
Also please tell me if you don't agree or don't understand some points! <3

PaulWoitaschek:
@happy-turtle 

I added a section on ktlint to the readme:
https://github.com/PaulWoitaschek/Voice/blob/main/README.md#ktlint

happy-turtle:
> @happy-turtle 
> 
> I added a section on ktlint to the readme:
> https://github.com/PaulWoitaschek/Voice/blob/main/README.md#ktlint

Thanks, that's helpful

app/src/main/kotlin/voice/app/injection/PrefsModule.kt
happy-turtle: I'm struggling with this one. Could you outline a bit more what you have in mind? I got as far as adding a new `AutoSleepTimerPrefAdapter` to the pref module. But I would think ideally this should rather be a more generic Adapter for any `Serializable` object?

```
@Serializable
data class AutoSleepTimerPrefs(val enabled: Boolean, val startTime: String, val endTime: String) {
  fun getStartTime(): LocalTime = LocalTime.parse(startTime)
  fun getEndTime(): LocalTime = LocalTime.parse(endTime)
}

internal object AutoSleepTimerPrefsAdapter : InternalPrefAdapter<AutoSleepTimerPrefs> {

  override fun get(
    key: String,
    prefs: SharedPreferences,
  ): AutoSleepTimerPrefs {
    val jsonString = prefs.getString(key, null)
    return if (jsonString != null) {
      try {
        Json.decodeFromString(jsonString)
      } catch (e: Exception) {
        getDefault()
      }
    } else {
      getDefault()
    }
  }

  override fun set(
    key: String,
    prefs: SharedPreferences,
    value: AutoSleepTimerPrefs,
    commit: Boolean,
  ) {
    val jsonString = Json.encodeToString(value)
    prefs.edit(commit = commit) { putString(key, jsonString) }
  }

  private fun getDefault() = AutoSleepTimerPrefs(
    enabled = false,
    startTime = LocalTime.of(22, 0).toString(),
    endTime = LocalTime.of(6, 0).toString(),
  )
}

```

PaulWoitaschek: Yes, that could work, but also shared preferences are kind-of deprecated anyways.
You could go with a DataStore, for this we already have a serializer ready:
```kotlin
  @Provides
  @Singleton
  @CurrentBook
  fun currentBook(factory: VoiceDataStoreFactory): DataStore<BookId?> {
    return factory.create(
      serializer = BookId.serializer().nullable,
      fileName = "currentBook",
      defaultValue = null,
    )
  }
```

playbackScreen/src/main/kotlin/voice/playbackScreen/BookPlayViewModel.kt
```diff
@@ -182,9 +189,27 @@ class BookPlayViewModel
         }
       }
     }
+    if (autoSleepTimerPref.value && !sleepTimer.sleepTimerActive()) {
```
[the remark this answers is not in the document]
happy-turtle: I definitely agree that this logic should be handled by the player instead. When trying to implement it there, I got a bit into recursive dependencies between the playback and sleepTimer modules though. The sleepTimer module, containing logic and UI at the same time, depends on the playback module. Do you have a suggestion on how we might trigger the sleepTimer from the player without creating a recursive dependency?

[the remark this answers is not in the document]
PaulWoitaschek: I thought about this again and this should be easily doable reactively by the SleepTimer itself.

It can listen for the playback state:

```kotlin
  init {
    scope.launch {
      playStateManager.flow.collect { playState ->
        if (playState == Playing) {
          activateIfInRange()
        }
      }
    }
  }
```

PaulWoitaschek:
@happy-turtle 
Would it be okay for you if I pushed some commits to your branch?

happy-turtle:
Yes, go ahead. I already pushed all my latest work.

strings/src/main/res/values-de/strings.xml
```diff
@@ -126,8 +126,11 @@
     <string name="sleep_timer.title">Schlaf gut! 🌙</string>
     <string name="sleep_timer.button.increment">Länger</string>
     <string name="sleep_timer.button.decrement">Kürzer</string>
+    <string name="auto_sleep_timer">Starte Schlummerfunktion automatisch</string>
```
PaulWoitaschek, line 129: Please don't add any != en localizations yet. These will be added to weblate and then pushed automatically

pref/src/main/kotlin/voice/pref/AutoSleepTimerPrefs.kt
```diff
@@ -0,0 +1,10 @@
+package voice.pref
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class AutoSleepTimerPrefs(
+  var enabled: Boolean,
+  var startTime: String,
```
PaulWoitaschek, line 8: I'd use `LocalTime` here and then create a custom serializer.
You can follow the `UriSerializer`  to see that in action
```kotlin
  @Serializable
  data class SelectFolderType(
    val uri:
    @Serializable(with = UriSerializer::class)
    Uri,
    val mode: Mode,
  ) : Compose {

    enum class Mode {
      Default,
      Onboarding,
    }
  }
```

gradle/libs.versions.toml
```diff
@@ -115,6 +116,7 @@ compose-typedNavigation = "com.kiwi.navigation-compose.typed:core:0.10.0"
 ktlint-core = { module = "com.pinterest.ktlint:ktlint-cli", version.ref = "ktlint-core" }
 ktlint-gradlePlugin = { module = "io.github.usefulness:ktlint-gradle-plugin", version.ref = "ktlint-gradlePlugin" }
 ktlint-compose = "io.nlopez.compose.rules:ktlint:0.4.22"
+datastore-core-android = { group = "androidx.datastore", name = "datastore-core-android", version.ref = "datastoreCoreAndroid" }
```
PaulWoitaschek, line 119: You should be able to use the already existing data store dependency:
```
datastore = "androidx.datastore:datastore-preferences:1.1.4"
```

(To use it, just call `implementation(libs.datastore)`)

playbackScreen/src/main/kotlin/voice/playbackScreen/BookPlayViewModel.kt
```diff
@@ -182,9 +187,32 @@ class BookPlayViewModel
         }
       }
     }
+    autoSleepTimerPref.data.map { prefs ->
```
PaulWoitaschek, line 190: This won't work like this.
Calling `map` transforms the flow, but without calling `collect()` at some point, the code inside map will never be executed.

Instead you usually use something like:
```kotlin
scope.launch {
  flow.collect { prefs->
    // logic here
  }
}
```

pref/src/main/kotlin/voice/pref/AutoSleepTimerPrefs.kt
```diff
@@ -0,0 +1,10 @@
+package voice.pref
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class AutoSleepTimerPrefs(
+  var enabled: Boolean,
```
PaulWoitaschek, line 7: Generally its a good practice to keep as much immutable as possible.
For example we collect this from a flow.
That means if you would now call

prefs.enabled = false

That change would never properly propagate.
I recommend to make these three properties `val`s to prevent accidental misusage

PaulWoitaschek:
Thanks @happy-turtle! Great stuff <3

PR #2738, opened 2025-04-09
Remove the sleep timer from the notification

PaulWoitaschek:


sandstorage:
Hey @PaulWoitaschek 
Thanks for this cool project!
I was a big user of this sleep button in the notification.
Do you have an insight on why it was removed?

Ram-Z:
> Hey @PaulWoitaschek Thanks for this cool project! I was a big user of this sleep button in the notification.

Same here. It was a very convenient way of setting the sleep timer without having to open the app and press a button at top of screen then bottom of screen.

btp:
I also use this feature and am disappointed it's gone. But thank you so much for developing this app!

PR #2824, opened 2025-05-18
fix: apply bottomSheet container color to items

FineFindus:
Fixes an issue, where the content of the bottomSheet would keep the default background color, instead of following the bottomSheet container color. This is especially noticeable when using stronger color schemes.

| Before                                                                                                                        	| After                                                                                                                             	|
|-------------------------------------------------------------------------------------------------------------------------------	|-----------------------------------------------------------------------------------------------------------------------------------	|
| Bottomsheet with default background color 	| Bottomsheet with bottomsheet background color 	|


PaulWoitaschek:
Thanks! Did you check on dark mode too?

FineFindus:
Yup, the issue exists on dark mode and is fixed by this ;)

PaulWoitaschek:
> Yup, the issue exists on dark mode and is fixed by this ;)

Awesome, thank you! 🎉

Issue #2265, opened 2024-01-21
Not possible to exit Chapter selection


littlesandra88:
### Overview

It is not possible to exit the chapter selection without selecting a chapter, and hence restarting the chapter.

It should be possible to enter the chapter selection without and then exit it with choosing a chapter. E.g. to see how many chapters a book have.

### How to reproduce

1. Open the App
2. Click on Play
3. Click Chapter
4. 

### Version

8.1.3

### Android Version

13

### Device Model

Pixel 6a

PaulWoitaschek:
Can't you just press the back button or touch somehwere outside the dialog?

littlesandra88:
On my phone atleast the chapter selector dialog fills the entire screen, so it is not possible to click outside. I don't see a back button. Is it suppose to be inside the chapter dialog?

PaulWoitaschek:
Could you post a screenshot?

Regarding the back button:
You phone either has a back button. Or you have gestures enabled (swiping from the very right or left of the screen) to close elements.

littlesandra88:
I didn't knew about the swipe gestures. Thanks =) Personally I prefer a visual Close button, but maybe that sort of thing is out dated?

Here are some screenshots =)

![signal-2024-01-21-172246](https://github.com/PaulWoitaschek/Voice/assets/2535545/d16c8ca9-64eb-4566-8509-5d58abbd0702)

![signal-2024-01-21-172242](https://github.com/PaulWoitaschek/Voice/assets/2535545/83f3c90b-69d3-4f88-950c-fb6bc9430134)

PR #2825, opened 2025-05-18
feat: redesign chapter selection

FineFindus:
Redesigns the chapter selection to use a bottom sheet and better highlight the current chapter. The increased space improves the display of long chapter titles, as well as allowing to show the time when each chapter starts.

Chapter bottom sheet

Closes: https://github.com/PaulWoitaschek/Voice/issues/2265

### DIFF
--- a/playbackScreen/src/main/kotlin/voice/playbackScreen/SelectChapterDialog.kt
+++ b/playbackScreen/src/main/kotlin/voice/playbackScreen/SelectChapterDialog.kt
@@ -27,7 +27,7 @@ internal fun SelectChapterDialog(
     onDismissRequest = { viewModel.dismissDialog() },
     content = {
       LazyColumn(
-        state = rememberLazyListState(initialFirstVisibleItemIndex = dialogState.selectedIndex ?: 0),
+        state = rememberLazyListState(initialFirstVisibleItemIndex = dialogState.selectedIndex?.minus(1)?.coerceAtLeast(0) ?: 0),
         content = {
           itemsIndexed(dialogState.chapters) { index, chapter ->
             ListItem(
--- a/playbackScreen/src/main/kotlin/voice/playbackScreen/SelectChapterDialog.kt
+++ b/playbackScreen/src/main/kotlin/voice/playbackScreen/SelectChapterDialog.kt
@@ -1,6 +1,7 @@
 package voice.playbackScreen
 
 import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.foundation.lazy.rememberLazyListState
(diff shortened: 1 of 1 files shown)


PaulWoitaschek:
Thanks! There is one bug: If the current chapter is not in the first 4-5 items, but for example on position 10, when opening the chapter list, the selected chapter is not visible.

playbackScreen/src/main/kotlin/voice/playbackScreen/SelectChapterDialog.kt
```diff
@@ -1,53 +1,65 @@
...
 ) {
-  AlertDialog(
+  ModalBottomSheet(
     onDismissRequest = { viewModel.dismissDialog() },
-    confirmButton = {},
-    text = {
+    content = {
       LazyColumn(
-        state = rememberLazyListState(initialFirstVisibleItemIndex = dialogState.selectedIndex ?: 0),
+        state = rememberLazyListState(initialFirstVisibleItemIndex = dialogState.selectedIndex?.minus(1)?.coerceAtLeast(0) ?: 0),
```
PaulWoitaschek, line 35: Thanks! There is one bug: If the current chapter is not in the first 4-5 items, but for example on position 10, when opening the chapter list, the selected chapter is not visible.

FineFindus: Hmm, I can't reproduce that, opened an audiobook, jumped 10 chapters forward, but the current chapter is still shown as the second visible item.

PaulWoitaschek: Can you screenshot?

FineFindus: After testing a bit, I think what's happening is that the LazyColumn always uses the maximum height of the ModalBottomSheet (i.e. screen height when it's fully expanded) and it's just moved into view when scrolling up.
As I rarely deal with Jetpack Compose, I'm a bit out of my depth here, maybe we can constrain the height of the column until the sheet is expanded? Or just fully expand it if the last chapter is selected?

PaulWoitaschek: I tested it futher with more chapters and the UX of the bottom sheet is really not made for more than a handful of items.

An alternative to really add more estate would be to move it to its own screen.
[ux.webm](https://github.com/user-attachments/assets/862574e8-d664-4f98-a2c7-0fd83317ab1a)

PaulWoitaschek:
I made two more changes: I added the chapter number (often times audiobook chapters are poorly named and every chapter is just called "Chapter".
I also changed the time calculation that it accounts for the previous files durations. (See testcase)

playbackScreen/src/main/kotlin/voice/playbackScreen/SelectChapterDialog.kt
FineFindus: This seems to be a bug in the compose implementation of bottomsheets (I tested it with a 'normal' bottomsheet and a recyclerview, it works flawlessly). 

I'm personally not a huge fan of adding a new screen just for the chapters, since that causes the chapters to feel somewhat disconnected from the player. Also in that case we could also just always expand the sheet :)

PaulWoitaschek: Solved by setting skipPartiallyExpanded = true.

PaulWoitaschek (approved):

PaulWoitaschek:
Thank you @FineFindus! Really great <3

PR #2913, opened 2025-06-29
fix: Fix a bug on Pixel Playback

### LINKED ISSUE #2902: Audio cuts off - Android 16 / Pixel 8
### Overview

I wrote it by mail but I'll leave it here as well!

After the update to Android 16 the audio is cutting off after about a minutes when I switch off the screen. From that point on the app is in a weird state in which the time slider seems to be stuck around where I switched off the screen, the play/pause button is showing pause without me being able to interact with it and the sleep timer continues to count down.

Most likely some more aggressive battery saving got introduced with the update but I could not immediately find anything to tune the settings for the app.

Glad to provide more information or test things if needed.

Thanks for the app and keep on the great work 🙂

### How to reproduce

Play an audiobook, turn off the screen and wait a bit.

### Version

8.6.1

### Android Version

Android 16

### Device Model

Pixel 8

PaulWoitaschek:
Audio Offload did lead to playback failures on Pixel devices. If this doesn't even work on Pixel devices, this is not stable enough.

Fixes #2902

### DIFF
--- a/playback/src/main/kotlin/voice/playback/di/PlaybackModule.kt
+++ b/playback/src/main/kotlin/voice/playback/di/PlaybackModule.kt
@@ -1,11 +1,9 @@
 package voice.playback.di
 
 import android.content.Context
-import android.os.Build
 import androidx.media3.common.AudioAttributes
 import androidx.media3.common.C
 import androidx.media3.common.Player
-import androidx.media3.common.TrackSelectionParameters.AudioOffloadPreferences
 import androidx.media3.datasource.DefaultDataSource
 import androidx.media3.exoplayer.ExoPlayer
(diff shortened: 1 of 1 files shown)


FineFindus:
Sad to see this removed :(. To add a counter data point, I have three different Pixel devices running without any issues.

PaulWoitaschek:
Yep agreed, super sad. 
But right now it's not worth it, it already caused problems for all the latest Samsung devices. 

It's also too specific to create a setting for this. 

And we have no way to reliably detect which devices cause troubles and which not. 
We can revisit with Android 17.

FineFindus:
Sounds good to me. I've added a TODO for roughly next year :)

PR #3134, opened 2025-09-21
3132 fts query

### LINKED ISSUE #3132: Search seems to rely on a strange query preparation, that does not seem right?!
### Overview

I stumbled over this while trying to search for the new `series` and `part` fields recently added, e.g `Benjamin Blümchen 124`

When searching a book, the search query is rewritten prepared, to use the `MATCH` keyword in an sqlite full text search using `FTS4` here:

https://github.com/PaulWoitaschek/Voice/blob/main/core/search/src/main/kotlin/voice/core/search/BookSearch.kt

This is the relevant code snippet:

```kotlin
  suspend fun search(query: String): List<Book> {
    return dao.search(
      buildString {
        append("\"")
        append('*')
        append(query.trim().replace("\"", "\"\""))
        append('*')
        append("\"")
      },
    ).mapNotNull { repo.get(it) }
  }
```

So searching for `Harry Potter 1` will become `"*Harry Potter 1*"`.

After doing a bit of research, I think a few things are strange:

- Enclosing the whole query in `"` (double quotes) results in searching for an `exact phrase` query, which should not work for wildcards at all (I still wonder why this still seems to work somehow, probably the asterisks and quotes are just ignored?)
- Prepending and appending an asterisk (`*`) seems strange, since sqlite only supports prefix matching (e.g. `hel* worl*`) - a leading asterisk `*arry *otter` seems not to be supported at all (is this an Android specific extension?)
- Replacing `"` (double quotes) within the query would search for the real existence of double quotes in the target string, which seems to have the same effect than removing it completely, because it does not even seem to be stored in the index

So I think this could be changed to something like this:

```kotlin
  suspend fun search(query: String): List<Book> {
    val searchString = buildString {
      append(
        query.trim()
          // replace all special chars except umlauts and other language specific letters
          .replace("[^\\p{L}0-9\\s]".toRegex(), " ")
          // replace all whitespaces with *
          .split("\\s+".toRegex()).joinToString("*")
      )
      append('*')
    }
    val results = dao.search(searchString)

    return results.mapNotNull { repo.get(it) }
  }
```

Now searching for `Harry Potter 1` will become `Harry*Potter*1*`.

`the wise man's fear 2.1` becomes `the*wise*man*s*fear*2*1*`, which should pretty likely result in what would be expected...

Could you elaborate on your decision on how to prepare the query for using it in the MATCH part?

### How to reproduce

Search for `Harry Potter 1`, when the test data contains 

### Version

Latest

### Android Version

13

### Device Model

Pixel 4a

sandreas:
Fixes #3132.

While the copilot pull request is a good approach, this one includes more sophisticated unit tests together with `BookFactory` adjustments to search for the new fields `genre`, `narrator`, `series` and `part`. The production code is basically the same.

@PaulWoitaschek 
A 2 minute review should already be enough to decide which one you'd prefer :-)

### DIFF
--- a/core/search/src/main/kotlin/voice/core/search/BookSearch.kt
+++ b/core/search/src/main/kotlin/voice/core/search/BookSearch.kt
@@ -4,41 +4,25 @@ import dev.zacsweers.metro.Inject
 import voice.core.data.Book
 import voice.core.data.repo.BookRepository
 import voice.core.data.repo.internals.dao.BookContentDao
+import voice.core.logging.api.Logger
 
 @Inject
 class BookSearch(
   private val dao: BookContentDao,
   private val repo: BookRepository,
 ) {
-
-  suspend fun search2(query: String): List<Book> {
-    val searchString = buildString {
+  suspend fun search(query: String): List<Book> {
+    val matchQuery = buildString {
       append(
         query.trim()
           // replace all special chars except umlauts and other language specific letters
           .replace("[^\\p{L}0-9\\s]".toRegex(), " ")
           // replace all whitespaces with *
-          .split("\\s+".toRegex()).joinToString("*"),
+          .split("\\s+".toRegex()).filter { it.isNotEmpty() }.joinToString("*"),
       )
       append('*')
     }
-    val results = dao.search(searchString)
-
-    return results.mapNotNull { repo.get(it) }
-  }
-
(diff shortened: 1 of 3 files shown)


PaulWoitaschek:
Thanks. In the future please do not merge without approval.

PR #3160, opened 2025-10-03
Fix a bug on adding multiple folders

PaulWoitaschek:
Recreate viewModel when uri or mode changes. Previously this was always re-using the same viewmodel with outdated parameters.

This is reproducible by:
- Going to the folder picker
- Selecting a folder A
- Going back and selecting another folder B

Now the first foler A is shown instead of B.

This PR fixes this.


sandreas (approved):
Code looks good, checked out and tested the fix, works like expected. Approved.

PaulWoitaschek:
> Code looks good, checked out and tested the fix, works like expected. Approved.

Thank you!

PR #3212, opened 2025-10-25
Add an opt-in for Analytics

PaulWoitaschek:
- By default opt-out of analytics
- Add a consent screen for analytics
- Track ScreenViews + Playback to find drop-off rates
- Track sleep timer usage to be able to prioritize development

### DIFF
--- a/core/playback/src/test/kotlin/voice/core/playback/player/VoicePlayerTest.kt
+++ b/core/playback/src/test/kotlin/voice/core/playback/player/VoicePlayerTest.kt
@@ -109,6 +109,7 @@ class VoicePlayerTest {
     mediaItemProvider = mediaItemProvider,
     volumeGain = mockk(relaxed = true),
     sleepTimer = mockk(relaxed = true),
+    analytics = mockk(relaxed = true),
   )
 
   @Test


sandreas:
Just to ensure I'm understanding this right: This is a feature to track user behavior if the user agrees? The reason is to analyze user behavior and improve the app. 

Not that I have something against it, but to me this feels a bit coming out of the blue, is there a specific reason for this? And who will have access to this data?

PaulWoitaschek:
Correct. Only me will have access and it's anonymized.

I need user data to make decisions. For example: to decide to put a focus on bookmarks, or rather simplify them, I need to understand if it's a niche feature or something that's a core feature.

PaulWoitaschek:
And more importantly:
I have the feeling that our onboarding and the folder modes are highly confusing to users. 

I want to find out the distribution of folder modes. 
If for example one is a super edge case, I want to think about removing it to reduce complexity.

sandreas:
>  If for example one is a super edge case, I want to think about removing it to reduce complexity.

Ok... the review might take a day or two, I'm pretty busy atm - I'll do my best to push this through.

PaulWoitaschek:
@sandreas 

I've merged it as I want to bring it live but would love a review anyways :)

sandreas:
@PaulWoitaschek 
Ok, started the review. Code looks good so far - but somehow the analytics don't show up when I build it on my dev device. Do I need to enable something in the manifest? Something like this?

```
<application>
    <meta-data
      android:name="firebase_analytics_collection_enabled"
      android:value="true" />
  </application>
```

Or am I doing something wrong?

sandreas:
@PaulWoitaschek 

I'm sorry to ask you again: Did you read my message?

PaulWoitaschek:
Ah sorry I missed this! 

I think it's not even easily possible to test it as one would need the Google service account credentials. 
Ill think about simplifying the general setup though.

sandreas:
> I think it's not even easily possible to test it as one would need the Google service account credentials.

Ok at least it compiles :-)

While I'm not a huge fan of "only, works on my system" features, in this case I'm not going to complain since A: It is opt-in and B: I probably would not find the time to look into the captured data.

So if you'd like to, you could add an sqlite storage provider, which maybe would help on mainly offline devices and sync later or a simple file-log or debug-log provider spitting messages out instead of structured logging... Just a thought.

PaulWoitaschek:
I did think about it a bit.
https://github.com/PaulWoitaschek/Voice/blob/4a1df7f21d329a686d689a9b72b6e6add24841db/app/build.gradle.kts#L102

Here you could set the value to "true". That would still mean that it's not the real analytics but the NoOp implementation, but it would show the consent screen.

It's not really "only works on my system" but rather "only works if you pull in an account".

PR #3383, opened 2026-03-07
Extend sleep timer reset window and add media button support

phreeza:
This PR extends the sleep timer's reset window to 2 minutes and allows the timer to be reset by pressing the headphone play button during this window.

The longer window is personal preference happy to change that back if you prefer.


phreeza:
I made a build and installed it on my phone, it works.

luzpaz:
@phreeza are you planning to address the review feedback ?

phreeza:
Not sure, I might if I find some time but for the time being I'm just using
it as is with and adb build on my phone.

PR #3437, opened 2026-04-10
Fix(bug): empty bottom sheet on book long-press.

methos5000:
BookOverviewScope was an annotation class, which Metro 0.13.2 does not recognize as a valid aggregation scope. As a result, all ViewModels in the BookOverview graph were unscoped — new instances on every access.

This caused the bottom sheet to appear empty: bookSelected() populated one BottomSheetViewModel instance, but the recomposition triggered by showBottomSheet=true accessed a fresh instance with empty state.

Convert BookOverviewScope to an abstract class (Metro's required pattern for aggregation scopes) and apply @SingleIn(BookOverviewScope::class) to all seven ViewModels in the feature.


PaulWoitaschek:
Yikes, same as https://github.com/PaulWoitaschek/Voice/commit/bdac8892f09b57517c5b704d6b46c9287bdc8bac

Good catch!

PR #3503, opened 2026-05-02
Feature/audiolog

joe42:
Hi!
Thank you for creating Voice! It's an excellent app.
I was missing an audio log similar to what the audible app has, so I added a feature for that and would like to share it.
The code is a bit hacked together and I am still testing it. Seems to work so far, though.

Feel free to merge change and use however you like.

I added an audiolog which                                                                                                                                                            
   - Adds an **Audiolog** screen (alongside the existing Bookmarks screen) that automatically records entries when playback is paused, before short/chapter skips, before  manual position changes, and when the sleep timer fires
   - Short-skip entries are ignored within a 30-second window to avoid noise 
   - Adds a Room schema migration (version 59) for the new `setByAudiolog` column

   ## Changes

   - New `AudiologRecorder` in `core/playback` wires into existing playback events
   - `AudiologScreen` composable split out from the bookmark screen, with its own edit dialog (`EditAudiologDialog`)
   - Add button removed on the audiolog screen (entries are auto-created only)
   - Bookmark and audiolog screens refactored into separate composables
 
image
image


nearapogee:
This would be crazy helpful.

joe42:
I rebased and added the context of where the position is skipping to, like "setting position to HH:MM:SS" or "skipping to HH:HH:SS". Not entirely sure if that is useful, but migh be easier to understand. I think it's not intuitive that the audo log position is the one before the position changed. Maybe that helps or maybe something like renaming it from skipping to "position before skipping" is easier to read, though either version is more verbose.

Otherwise it's working fine. I find it very useful since
1. it helps me find the position I stopped listening in case I  used sleep mode
2. it helps me change the position  back if I set it to the wrong position by accident *

*) My navigation on the lock screen is too small, my fingers are big and my screen protector makes the screen less sensitive all of which combines to making rewinding so hard I often accidentally set the position to 0 by accident.

It's a trade off between an additional feature vs additional complexity though.
Maybe it's best to hide it behind a feature flag and let user's enable it via the settings in case you want to include it into main.

I'm happy using my debug version myself. If you decide it's useful let me know and I'll try to polish it some.

tashilapathum:
I also needed this feature, thank you @joe42!

PaulWoitaschek:
Thank you for this contribution! 

I am also planning with this feature for quite a while. 

There are some technical decisions made in this PR which I would change. 

But before going into the details: are you an Android developer or has this been vibe coded? 
Depending on the answer, it would either give technical guidance and shape it myself. 
If this was human effort I would built on top of it.

joe42:
Hi Paul,
I'm happy to hear that!
I  am not a Android developer nor have I coded with Kotlin before. The entire development was heavily AI assisted. 
My goal is to make a small contribution and start learning how to develop Android apps myself.

For me either option sounds great. If the changes find their way in the main branch that's already a bonus.

I felt relatively comfortable directing and reviewing the changes. I think the things I struggled most with is the UI design since I'm more of a backend/data developer. So I copied most of that and modified things from there. I also let the AI create the migration scripts and only reviewed them on a superficial level, though I think I will be able to pick up the mechanism behind that fairly easily.

If it helps I'd definitely enjoy pouring some more effort into the project while getting more familiar with Android dev and Kotlin.
So if you need some help with a smaller task let me know any time. And if you want to put in the extra effort of explaining how I can improve the PR I would of course give rewriting the feature with the help of your guidance a try. I think that would be a great way to learn for me. 
It's your project, so totally up to what you think is best.

joe42:
----

joe42:
Oh, no offense! 
I thought I added the tables and had just called them bookmark2 and content2 instead of naming them something like audiolog and audiolog_content. But now I remember that I only added a new column.

PaulWoitaschek:
Ok, thanks!

I'll be closing this for now until we have a good UI / UX concept.

What I was thinking of is some form of a vertical timeline, or a joined bookmark / audiologs screen where you'd have filter tags to see and hide bookmarks / audio-log entries.

Some technical feedback:
- Generally I would separate the data models here and have bookmarks and audio-logs as separate concepts.
- The AudioLog recorder should not receive raw strings "setting position". Instead it should receive and store a form of an enum 

```kotlin
enum class AudioEvent {
  Seek, SkipForward, ...
}
```

and store these in the database.

Then in the UI layer you'd render these enum entries.

Also I'd likely limit these, as they would quickly become many.

joe42:
> Some technical feedback:
> 
>     * Generally I would separate the data models here and have bookmarks and audio-logs as separate concepts.
> 
>     * The AudioLog recorder should not receive raw strings "setting position". Instead it should receive and store a form of an enum and store these in the database.

Yes, that makes sense.

> What I was thinking of is some form of a vertical timeline, or a joined bookmark / audiologs screen where you'd have filter tags to see and hide bookmarks / audio-log entries.

I did start out with only using the bookmark dialog at first which worked out well. 
I think I actually only split that out to isolate the components on a technical level and realized afterwards that with the audio log I don't need the bookmarks anymore myself.
Handling both in a single dialog sounds like a good idea. Is it worth factoring out the audio log model if the interface should be the same for bookmarks and audio logs? I guess it gives some more flexibility if the features diverge later on.

PR #3505, opened 2026-05-04
fix: Crop long title in AppBar (fixes #2853)

### LINKED ISSUE #2853: feat: Improving the long title
Crop the long scrolling title on the right as it is on the left, see image:

Image

ikanakova:
## Description
Fixes the issue where the long title on the right side.

## Changes
- Added `.fillMaxWidth()` to use available space
- Added `.clipToBounds()` to crop text at boundaries
- Kept `.basicMarquee()` for smooth scrolling animation

## Fixes
Fixes #2853 #3163

### DIFF
--- a/features/playbackScreen/src/main/kotlin/voice/features/playbackScreen/view/AppBarTitle.kt
+++ b/features/playbackScreen/src/main/kotlin/voice/features/playbackScreen/view/AppBarTitle.kt
@@ -1,14 +1,19 @@
 package voice.features.playbackScreen.view
 
 import androidx.compose.foundation.basicMarquee
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.material3.Text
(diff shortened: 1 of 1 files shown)


PaulWoitaschek:
Thanks for the PR!
It's still cropped on the left side

image

ikanakova:
Adding end padding seems to have solved the problem - tested on a Nokia 6.1 (LineageOS 22.2, Android 15):
Screenshot_20260511-182522_Voice  Debug

PaulWoitaschek:
I think the better way would be to not crop it.
By default start with a margin in the left and then let marquee flow with full width.

PaulWoitaschek:
It's still cropped. Not sure if this can be achieved without hacking things from fresh (which won't be maintainable either). Closing this for now, thanks for your effort!

PR #3526, opened 2026-05-16
Speed up media scan

joe42:
## Summary

Reduces SAF (Storage Access Framework) traffic during media scans by reusing a single per-book directory walk and eliminating a redundant ExoPlayer analyze call for newly discovered books.

- Walk each candidate book exactly once into a `List<AudioFile>`, then reuse that list for the permission probe, ordering, and chapter parsing (previously three separate walks per book).
- Replace `sortedBy { it.audioFileCount() }` with `sortedBy { it.second.size }`. Kotlin's `sortedBy` does not cache its selector, so the old version re-walked each book's SAF tree on every comparison (`O(log N)` walks per book).
- `ChapterParser.parse` now takes the pre-walked list and returns `ChapterParseResult(chapters, firstFileMetadata)`. `BookParser.parseAndStore` accepts the optional first-chapter metadata so newly discovered books skip a duplicate `MediaAnalyzer.analyze` call. Null path preserves prior behavior.

### DIFF
--- a/core/scanner/src/main/kotlin/voice/core/scanner/BookParser.kt
+++ b/core/scanner/src/main/kotlin/voice/core/scanner/BookParser.kt
@@ -24,11 +24,12 @@ internal class BookParser(
   suspend fun parseAndStore(
     chapters: List<Chapter>,
     file: CachedDocumentFile,
+    firstChapterMetadata: Metadata? = null,
   ): BookContent {
     val id = BookId(file.uri)
     return contentRepo.getOrPut(id) {
-      val uri = chapters.first().id.toUri()
-      val analyzed = mediaAnalyzer.analyze(fileFactory.create(uri))
+      val analyzed = firstChapterMetadata
+        ?: mediaAnalyzer.analyze(fileFactory.create(chapters.first().id.toUri()))
       parse(chapters, id, analyzed, file)
     }
   }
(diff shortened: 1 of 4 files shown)


PaulWoitaschek:
Interesting! How does this change the duration for your library?

joe42:
Thanks!
I didn't debug it but it looks like a significant change for me.
I added the change after the scan started taking annoyingly long one time due to my library growing quite large and with the change in place it seems to be much more snappy. I tested it a bit and would guesstimate it to be about a 4x speedup.

PaulWoitaschek:
Thanks for the PR!

I'm skeptical that this can really create a 4x speed-up.

`CachedDocumentTree` already is lazy and caches the SAF accesses.
So all optimizations around that (like audioFileCount) should not have a real impact on scanning performance.

There's one case though that's worth optimizing, and that is calling analyze one additional time per book.

So a case like this:

- Book1
  - file1.mp3
  - file2.mp3
  - file3.mp3
 
Would lead to 4 instead of 3 analyze calls. I've handle that change in #3528

The maximum effect this could have is if someone had always exactly 1 file per folder, in this case this should lead to a 2x speed-up.

LMK if you have a different understanding!

joe42:
Good point, you're right.
I think just adding https://github.com/PaulWoitaschek/Voice/pull/3528 would be great. 

Happy to close this PR in favor of #3528.

Issue #3062, opened 2025-09-03
m4b chapters not showing


mpetuska:
### Overview

This m4b audiobook has chapter metadata (verified via smart audiobook player app), yet Voice does not show them (or any chapter selection for that matter).

https://drive.proton.me/urls/R1QT8SR8QW#kKksYGCoFG5P

### How to reproduce

Download and open this audiobook in voice
https://drive.proton.me/urls/R1QT8SR8QW#kKksYGCoFG5P

### Version

8.6.2

### Android Version

15

### Device Model

Pixel Fold

PaulWoitaschek:
@sandreas 
This reproduces to me. Could you have a look if you can fix this?

You can validate that by moving that file here:
https://github.com/PaulWoitaschek/Voice/tree/57449df58b003a8e37ec59109ed6c9e76cb19df7/core/scanner/src/test/resources

And then in `MediaAnalyzerTest`:

```kotlin
@Test
fun test() {
  parse("robert.m4b")!!.chapters.shouldHaveAtLeastSize(2)
}
```

An ffprobe shows 42 chapters. Strangely though we do detect the very first chapter here:
```
[MarkData(startMs=0, name=Opening Credits)] 
```

sandreas:
@PaulWoitaschek I'll take a look. May take some time.

PaulWoitaschek:
Maybe you can also convince the mediax people to do chapter parsing upstream 😁

sandreas:
> Maybe you can also convince the mediax people to do chapter parsing upstream 😁

I try my very best. Pretty likely the file is bogus or the chapter names contain invalid or very exotic chars (see dump below) - but I'll take a closer look soon.

```json
{
"chapters": [
      {
        "start": 2482236722,
        "title": "r 2\u0000\u0000\u0000\u0006�q��\tChapter 3\u0000\u0000\u0000\u000b�f��\tChapter 4\u0000\u0000\u0000\u000f��l\u0010\tChapter
5\u0000\u0000\u0000\u0012�7�\tChapter 6\u0000\u0000\u0000\u0016\u001e\u0006\tChapter 7\u0000\u0000\u0000\u0019q��p"
      },
      {
        "start": 1107001740,
        "length": 1215799097,
        "title": "8\u0000\u0000\u0000\u001c`w��\tChapter 9\u0000\u0000\u0000\u001f3\u001et \nChap"
      },
      {"...": "..."}
   ]
}
```

[...]

PR #3608, opened 2026-06-15
Fix MP4 chapter extraction for chapter tracks stored in a single chunk

eajaykumar:
While listening to a Harry Potter Atmos audiobook, I noticed that chapters were not being detected by Voice, even though the file contained valid chapter information.
After investigating the file structure, I found that the audiobook stores chapters in a dedicated chapter track referenced through the `chap` track atom. Unlike files handled by the current implementation, all chapter samples were stored in a single chunk instead of one chunk per chapter.

The existing parser assumes a direct relationship between chunk offsets and chapter entries, which caused chapter extraction to fail for this file.

This PR updates the MP4 chapter parser to correctly handle chapter tracks that contain multiple chapter samples within a single chunk.
**Changes**
**Track identification**
Added `TkhdVisitor` to collect track IDs from `tkhd` atoms.
Track metadata is now associated using the actual track ID instead of assuming `trackId - 1` maps to the correct track.
**Sample parsing**
Added `StszVisitor` to read sample sizes from `stsz` atoms.
Chapter names are now read from individual samples rather than assuming each chunk contains exactly one chapter.
**Chunk offset support**
Added `Co64Visitor` to support MP4 files using 64-bit chunk offsets (`co64` atoms).
**Parser updates**
updated `Mp4ChpaterExtractorOutput` to store track IDs and sample sizes.
Updated `ChapterTrackProcessor` to correctly process chapter tracks where multiple chapter samples are stored in a single chunk.
Updated `Mp4BoxParser` to support large atom payloads and stop parsing once `mdat` is reached.
**Testing**
Tested with:

A Harry Potter Atmos audiobook that previously showed no chapters.
Standard M4B audiobooks with existing chapter support.
Chapters are detected correctly in both cases.
 **Before** 
Screenshot_20260615_204304_Voice
**After**
Screenshot_20260615_204205_Voice  Debug
Screenshot_20260615_204209_Voice  Debug
**Normal M4B File**
Screenshot_20260615_204320_Voice

### DIFF
--- a/core/data/api/src/main/kotlin/voice/core/data/Chapter.kt
+++ b/core/data/api/src/main/kotlin/voice/core/data/Chapter.kt
@@ -12,7 +12,6 @@ import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
 import kotlinx.serialization.descriptors.SerialDescriptor
 import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
-import voice.core.logging.api.Logger
 import java.time.Instant
 
 @Entity(tableName = "chapters2")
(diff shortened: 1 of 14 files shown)


PaulWoitaschek (commented):
I am a bit hesitant with merging this. 

I am afraid that it will break the parsing of other media. It also misses file test cases. 

Ideally media 3 should be released any moment with official support for chapter parsing. 

The alternative could be to feature-flag your code and send an analytic event and get some real world data. "How many percentage of files parse with chapters" 
We might need that anyways when the official support comes and run it as an ab-test.

eajaykumar:
@PaulWoitaschek That makes sense,i just changed the logic of chapters parsing so it requires lot of testing to make sure it working,its good to lets wait for media3 support to land as its better anyway and thanks for this app

PaulWoitaschek:
Thanks for the contribution!

Sadly closing this as discussed in https://github.com/PaulWoitaschek/Voice/issues/3062

PR #3631, opened 2026-06-20
fix album cover cropping when chapters are displayed

eajaykumar:
improve the playback screen layout when chapters are available.

When chapter navigation is visible, the layout reduced the available height for the cover, causing it to be center cropped and parts of the artwork to be cut off.

ContentScale.Fit for cover artwork rendering in fixed aspect ratio
MediumTopAppBar for the tiltle name as so it dont take lot of space in playback screen.

The cover artwork remains square and is no longer cropped differently when chapter navigation is shown even when chapters takes 3 lines and Title name looks fine.

### DIFF
--- a/features/playbackScreen/src/main/kotlin/voice/features/playbackScreen/view/BookPlayAppBar.kt
+++ b/features/playbackScreen/src/main/kotlin/voice/features/playbackScreen/view/BookPlayAppBar.kt
@@ -8,6 +8,7 @@ import androidx.compose.foundation.layout.size
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.material3.LargeTopAppBar
+import androidx.compose.material3.MediumTopAppBar
 import androidx.compose.material3.TopAppBar
 import androidx.compose.material3.ripple
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


eajaykumar:
**Before**
Screenshot_20260615_204205_Voice  Debug
**After**
Screenshot_20260617_220758_Voice  Debug

PaulWoitaschek:
This works for some cases, but does not work as a general solution. 
This will cause empty spaces in covers, esp on non small devices. 

I believe cropping is better even though for some covers this means you won't see the whole cover. 

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
    <th>After</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
    <td></td>
  </tr>
</table>

eajaykumar:
@PaulWoitaschek i get it, making one device look better can make another one worse. we need something that adapts to all screens to make it better. i tried BoxWithConstraints but it also same result.So it doesn't seem like a general solution either.

eajaykumar:
UI adaptation is very complex topic as it takes lots of effect to fine tune each device screen but to implement smooth UI for all devices need to categorize small, large, tablets, foldables, and landscape.its intensive i just leave it for simple solution you implemented.

PR #3658, opened 2026-07-09
Feature: Add lockscreen seeking toggle

DevOelgaard:
# Description
Adds a setting to enable/disable seeking via the lockscreen media session timeline cursor. By default, seeking on the lockscreen is disabled to prevent accidental seeks when the phone is in a pocket or when the screen is accidentally touched.

This resolves the feature request to make the lockscreen seek bar non-actionable or at least toggleable via a setting.

Closes #3229
Closes #3589

-----
# Screenshot
e43eb9a7-f7bb-4e9f-98dc-e3622dfb09db

### DIFF
--- a/core/data/api/src/main/kotlin/voice/core/data/store/StoreQualifiers.kt
+++ b/core/data/api/src/main/kotlin/voice/core/data/store/StoreQualifiers.kt
@@ -46,3 +46,6 @@ public annotation class DeveloperMenuUnlockedStore
 
 @Qualifier
 public annotation class FeatureFlagOverridesStore
+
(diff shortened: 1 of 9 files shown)


elazarcoh:
@DevOelgaard thank you, hope to see it merged!

I would also like to disable it in the notification scroll, even if not in lockscreen, would you consider adding it to this PR as well?

DevOelgaard:
@elazarcoh that's a good idea! I will wait and see if this pr gets merged, if it does I will add it. I'm not inclined on spending time on this if the pr goes stale.

PR #3689, opened 2026-08-01
Customizable buttons to the playback notification

edoput:
Thanks for the nice project, it's a cute app that does what I need.

The sleeping timer and other commands are only accessible from the main screen, this PR introduces an additional button to the media player notification and makes the three buttons configurable. This allows me to turn on/off the sleeping timer or insert a bookmark directly from the media player (notification and lockscreen).

The following screenshot shows the additional button on the bottom-right of the notification. All clickable buttons in this notification except "this phone" and "play/pause" are now provided by the app instead of the framework.

Media player screenshot

This is my first android app contribution and while I am a developer I delegate the implementation to Claude and spent time doing user-acceptance in the emulator. I am now daily-driving this and will report/update the PR.

During my back and forth with Claude we also added two items to AGENTS.md . The first documents the practice of using the ViewModel to test screens. The second is about relying on the compiler to check exhaustiveness of code instead of checking each value of an enum. While the first may be useful the second is just nice to have.

### DIFF
--- a/core/data/api/src/main/kotlin/voice/core/data/notification/MediaNotificationPreferences.kt
+++ b/core/data/api/src/main/kotlin/voice/core/data/notification/MediaNotificationPreferences.kt
@@ -1,0 +1,30 @@
+package voice.core.data.notification
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+public enum class NotificationAction {
+  REWIND,
+  FAST_FORWARD,
+  NEXT_CHAPTER,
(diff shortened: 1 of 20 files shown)


luzpaz:
This is awesome. But it may be more strategic to split this PR? 

The changes you made in:  
> During my back and forth with Claude we also added two items to AGENTS.md . 

could be in a separate PR since they aren't directly relevant to the main feature you're proposing.

PR #3705, opened 2026-08-07
play large mp4 files via the platform MediaExtractor

ynotzort:
this is a follow up pr for https://github.com/PaulWoitaschek/Voice/pull/3704
it actually fixes the playback of large m4b files.

after I fixed the import of huge m4b files locally, I noticed that when they are played I get an OOM, because of the heapgrowthlimit. 
the actual issue is I think in the media3 implementation, not sure why they need almost 300mb heap for a 3gb audiobook file.

I tried using android.media.MediaParser but I get the same issue there. android.media.MediaExtractor however works, but needs a bit of extra work, thats what the PlatformMediaExtractor class is for. The Mp4AwareMediaSourceFactory is just there for routing the mp4 containers (those are the ones I had so far trouble with), the other filetypes are left untouched.

Note: 
- getCurrentInputPosition() is a bit hacky, we have no access to the MediaExtractor internals, so we just do the running sum of sample sizes here.
- Mp4AwareMediaSourceFactory needs to reapply the clipping config, otherwise chapters are not working properly.

tested it locally on my device with the problematic file.

used the following to generate a test m4b file:

```bash
#!/usr/bin/env bash
# Generate a long m4b for testing media3's per-sample sample-table allocation.
#
# The heap cost in Mp4Extractor is driven by the *sample count*, not by the file size:
# 32 bytes per audio frame (offsets 8 + timestamps 8 + sizes 4 + flags 4, plus 8 for
# accumulatedSampleSizes). AAC-LC at 44.1 kHz produces 43.07 frames/s = 155,040 frames per
# hour, so an hour costs 4.96 MB of java heap regardless of bitrate. Encoding silence at a
# low bitrate therefore gives a file that is byte-for-byte small but has exactly the sample
# table that triggers the OOM.
#
#   8.39M samples / 54.1 h  is where the arrays alone reach the 256 MB heapgrowthlimit
#   8.53M samples / 55 h    ~ Chrysalis Books 1-3, the file this was found on (8,479,389)
#   60 h and up             leaves headroom for the rest of the app, a safer reproducer
#
# Usage: ./make-test-m4b.sh [hours] [chapters] [output]
#        ./make-test-m4b.sh 55 433 test-55h.m4b
#
# Env:
#   FASTSTART=0   leave mdat before moov (default 1, moov first like the real file)
#   SOURCE=noise  encode noise instead of silence (silence compresses to ~1.3 kbps)
#   CODEC=alac    lossless, ~88 KiB/s on noise, for growing mdat quickly
#   BITRATE=96k   aac only
#
# To reproduce the >2 GB mdat integer overflow in Mp4BoxParser.parseBoxes instead, the mdat
# has to be both huge and *before* moov:
#
#   SOURCE=noise CODEC=alac FASTSTART=0 ./make-test-m4b.sh 7 20 test-2gb-mdat.m4b
set -euo pipefail

HOURS=${1:-55}
CHAPTERS=${2:-433}
OUT=${3:-test-${HOURS}h.m4b}
BITRATE=${BITRATE:-8k}
SOURCE=${SOURCE:-silence}
CODEC=${CODEC:-aac}
FASTSTART=${FASTSTART:-1}

SECONDS_TOTAL=$(python3 -c "print(int($HOURS * 3600))")
WORK=$(mktemp -d)
trap 'rm -rf "$WORK"' EXIT

case "$SOURCE" in
  silence) INPUT="anullsrc=r=44100:cl=mono" ;;
  noise) INPUT="anoisesrc=r=44100:c=pink:a=0.5" ;;
  *)
    echo "SOURCE must be 'silence' or 'noise'" >&2
    exit 1
    ;;
esac

python3 - "$SECONDS_TOTAL" "$CHAPTERS" > "$WORK/chapters.txt" <<'PY'
import sys
total_ms = int(sys.argv[1]) * 1000
count = int(sys.argv[2])
print(";FFMETADATA1")
print("title=Test Book")
print("artist=Test Author")
print("album=Test Book")
for i in range(count):
    start = total_ms * i // count
    end = total_ms * (i + 1) // count
    print("[CHAPTER]")
    print("TIMEBASE=1/1000")
    print(f"START={start}")
    print(f"END={end}")
    print(f"title=Chapter {i + 1}")
PY

MOVFLAGS=(-movflags "+faststart")
[ "$FASTSTART" = "1" ] || MOVFLAGS=()

CODEC_ARGS=(-c:a "$CODEC")
[ "$CODEC" = "aac" ] && CODEC_ARGS+=(-b:a "$BITRATE")

echo "encoding ${HOURS}h / ${CHAPTERS} chapters as ${CODEC} (${SOURCE}) -> $OUT"
ffmpeg -hide_banner -loglevel warning -stats \
  -f lavfi -i "$INPUT" \
  -f ffmetadata -i "$WORK/chapters.txt" \
  -map 0:a -map_metadata 1 \
  -t "$SECONDS_TOTAL" \
  "${CODEC_ARGS[@]}" \
  "${MOVFLAGS[@]}" \
  -f ipod -brand "M4B " \
  -y "$OUT"

ls -lh "$OUT"
```

### DIFF
--- a/core/playback/src/main/kotlin/voice/core/playback/di/PlaybackModule.kt
+++ b/core/playback/src/main/kotlin/voice/core/playback/di/PlaybackModule.kt
@@ -9,6 +9,7 @@ import androidx.media3.datasource.DefaultDataSource
 import androidx.media3.exoplayer.ExoPlayer
 import androidx.media3.exoplayer.source.DefaultMediaSourceFactory
 import androidx.media3.exoplayer.source.MediaSource
+import androidx.media3.exoplayer.source.ProgressiveMediaSource
 import androidx.media3.extractor.DefaultExtractorsFactory
 import androidx.media3.session.CommandButton
 import androidx.media3.session.MediaLibraryService
@@ -20,6 +21,8 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import voice.core.featureflag.FeatureFlag
 import voice.core.featureflag.Media3AudioOffloadFeatureFlagQualifier
+import voice.core.playback.mediasource.Mp4AwareMediaSourceFactory
+import voice.core.playback.mediasource.PlatformMediaExtractor
 import voice.core.playback.misc.VolumeGain
 import voice.core.playback.notification.MainActivityIntentProvider
 import voice.core.playback.player.DurationInconsistenciesUpdater
@@ -41,7 +44,10 @@ interface PlaybackModule {
     val dataSourceFactory = DefaultDataSource.Factory(context)
     val extractorsFactory = DefaultExtractorsFactory()
       .setConstantBitrateSeekingEnabled(true)
-    return DefaultMediaSourceFactory(dataSourceFactory, extractorsFactory)
+    return Mp4AwareMediaSourceFactory(
+      default = DefaultMediaSourceFactory(dataSourceFactory, extractorsFactory),
+      mp4 = ProgressiveMediaSource.Factory(dataSourceFactory, PlatformMediaExtractor.Factory(context)),
+    )
   }
 
   @Provides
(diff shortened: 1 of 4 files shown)


PaulWoitaschek:
Thanks! I believe this should be fixed in media3 directly. Can you raise an issue there?

PR #3720, opened 2026-08-15
Scan each book under an author folder only once

joe42:
Expanding an author directory in `MediaScanner` uses:

```kotlin
author.children.flatMap {
  author.children
}
```

The lambda ignores its parameter and re-emits the whole child list once per child.

For `Tolkien/` containing `[Hobbit, LotR, Silmarillion]` it yields:

```
[Hobbit, LotR, Silmarillion,
 Hobbit, LotR, Silmarillion,
 Hobbit, LotR, Silmarillion]
```

9 candidates instead of 3, so each book is scanned 3 times.


PaulWoitaschek:
Great finding! Could you add a test against regressions?

core/scanner/src/test/kotlin/voice/core/scanner/MediaScannerTest.kt
```diff
@@ -192,6 +193,21 @@ class MediaScannerTest {
...
+
+    audioFile(parent = audioBooks, "test.mp3")
+    audioFile(parent = audioBooks, "author1/test.mp3")
+    audioFile(parent = File(audioBooks, "author1/book1"), "c1.mp3")
+    audioFile(parent = File(audioBooks, "author1/book1"), "c2.mp3")
+    audioFile(parent = File(audioBooks, "author1/book2"), "a.mp3")
+
+    scan(FolderType.Author, audioBooks)
+
+    assertEquals(expected = 4, actual = scannedBooks.size)
```
Auric-Manteo, line 208: Four books: three below author1, plus the one directly in the root folder. 
Expanding an author folder used to re-emit its whole child list once per child,  so author1's three books were scanned three times each, giving 10 scans instead of 4.

core/scanner/src/test/kotlin/voice/core/scanner/MediaScannerTest.kt
```diff
@@ -294,4 +312,16 @@ class MediaScannerTest {
     val id: File,
     val chapters: List<File>,
   )
+
+  private class ScannedBooksRecordingRepo(
```
Auric-Manteo: Records the candidate list the scanner builds, so tests can assert each book is scanned once.

Auric-Manteo:
> Great finding! Could you add a test against regressions?

Happy to help!
I added a regression test.

Off topic - Thanks for adding in-chapter seeking on the lock screen, it's great!
I rebased the audiolog feature on top of it and have a working version now :)

